PR #400, opened 2013-03-11
Cancel menu mode on Ctrl-^ Esc

andersk:
Previously there was no way to get out of menu mode without taking some action (such as sending a keypress through).

Bikeshedding of UI details is welcome.

### DIFF
--- a/src/frontend/stmclient.cc
+++ b/src/frontend/stmclient.cc
@@ -286,6 +286,7 @@ bool STMClient::process_user_input( int fd )
 	} else if ( the_byte == '^' ) {
 	  /* Emulation sequence to type Ctrl-^ is Ctrl-^ ^ */
 	  network->get_current_state().push_back( Parser::UserByte( 0x1E ) );
+	} else if ( the_byte == 0x1b ) { /* Ctrl-^ Esc cancels menu mode */
 	} else {
 	  /* Ctrl-^ followed by anything other than . and ^ gets sent literally */
 	  network->get_current_state().push_back( Parser::UserByte( 0x1E ) );


cgull:
Now that mosh attempts to support both ssh-style and single-escape-character escapes, the water's even muddier on this one.  Imagine you're at a shell prompt, typing command<cr>~<esc> is actually pretty plausible, perhaps by way of an arrow key.

andersk:
@cgull, I’m not sure what you mean here; if by “ssh-style escapes” you mean `RET ~`, Mosh does not support that. (FWIW, personally I think ssh’s `RET ~` is terrible because you cannot invoke it without sending a spurious `RET` to the remote side.)

glensc:
ssh's `RET ~` is even more horrible, as `~` is dead-key in some international keyboards, so you may need to double press  the key to get literal `~`!

cgull:
@andersk, that's exactly what I mean.  I hadn't actually tried it till just now, but the code indeed works just fine, er, as intended.  Try `MOSH_ESCAPE_KEY='~' mosh some-host` (with sources from master) and be properly horrified.

cgull:
I have some WIP on the man pages to better document MOSH_ESCAPE_KEY, btw.

andersk:
Oh, so it does. But I’m still not quite sure what your comment means; are you suggesting that someone could accidentally type `RET ~ ESC` via `RET ~ UP`, since the encoding of `UP` begins with `ESC`? I’m not sure why that’s specific to `RET ~` though; it’s a general problem with `ESC`.

I’m having trouble thinking of an idiomatic replacement for `ESC`, though. Maybe `BACKSPACE`? (“I didn’t mean to type that `C-^`, so, `BACKSPACE`?”)

cgull:
Yeah, that the likelihood of accidents seems higher with `RET ~ ESC`, and that further confusion and/or accidental console input is possible since ESC is used both for ANSI special keys and multi-character interactive input.  I wanted to point that out and discuss it before pulling it; the best thing may be to use `ESC` anyway.  I doubt many people are using the SSH-style escape, because it's so undocumented and not obvious in the source code.

cgull:
Higher than with `Ctrl-^ ESC`, I mean.

PR #417, opened 2013-04-15
Provide an alternative mechanism for extracting the server IP address

pcc:
On some installations of ssh, overriding ProxyCommand is forbidden,
which means that the fake proxy won't work.   Instead, have mosh-server
report the local address in SSH_CONNECTION, which will be used if
the fake proxy doesn't report an address.


cgull:
There's a somewhat hackish way around this: mosh could do ssh $host 'echo MOSH SSHIP SSH_CONNECTION; mosh ...'

PR #420, opened 2013-05-10
Implement alternate screen

barosl:
Make it possible for Framebuffer to save and restore the current state of the screen. get_DEC_mode() in terminalfunctions.cc also needs to be patched to know the requested operation.


cgull:
The interaction between alt-screen and mouse modes is more complicated than that:  http://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Wheel-mice
We also need to deal with mosh suspend/resume.

PR #455, opened 2013-08-23
look for updated DNS when connection to server is lost

maltek:
I implemented issue 212:
Instead of resolving the host name in the mosh wrapper script and giving the mosh-client the resolved IPv4 address, I modified the script to pass down the host name, then changed the client to constantly re-resolve that host name when the connection is lost.

To do so, I modified Connection::hop_port() to make an asynchronous DNS request. The DNS requests are IPv4 only right now, but I used getaddrinfo, so changing it to support IPv6 should not be too hard.

Only GNU libc has an asynchronus getaddrinfo_a, so I had to use pthreads to make one.

Note also that I changed AC_CHECK_FUNCS to spit out an error message when a function isn't found, otherwise the missing getaddrinfo would just be ignored. I'm by far not an automake expert, so I'm not 100% sure this doesn't break anything.

I tested this patch on Linux, OS X and FreeBSD x86_64.

### DIFF
--- a/src/network/network.cc
+++ b/src/network/network.cc
@@ -268,7 +268,7 @@ Connection::DNSResolverAsync::Status Connection::DNSResolverAsync::try_start_sto
       if( !old_record_found ) {
         uint16_t old_port = addr.addr.port();
 
-        fatal_assert( result->ai_addrlen <= sizeof( remote_addr.addr ) );
+        fatal_assert( result->ai_addrlen <= sizeof( addr.addr ) );
         addr.len = result->ai_addrlen;
         memcpy( &addr.addr.sa, result->ai_addr, addr.len );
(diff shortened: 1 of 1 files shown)


cgull:
Looking at `athena.dialup.mit.edu`, I notice something interesting:  That hostname is actually a short-lived, often-changing CNAME pointing to a long-lived A record, with an apparently stable name referring to a specific machine.  So pretty clearly Athena is load-balancing via the CNAME.  Seems to me that for **this particular host**, a nice answer would simply be to have mosh resolve through any CNAME/DNAME records to an actual A/AAAA record, then remember and reuse that hostname.  I don't know how widely applicable this solution would be, though.

Were these records set up this way in 2013, or is this an improvement over whatever was in place then?

Issue #431, opened 2013-05-29
Having a .moshrc (equivalent to .ssh/config) file to be able to specify per-host configs


lemoinem:
Hi,

I think this would go a long way improving the usability of mosh to be able to have a .ssh/config equivalent.

Unfortunately, ssh complains about unknown directives in .ssh/config so we will need a new file (I also think it's good practice to do so anyway).

The format could be simple (.ssh/config sounds a pretty good example: easy to write, easy to parse).

Useful configs to include: `hostname`, `port`, `mosh-server`, eventually `ssh` and `command` (in some weird case). Later: enable/disable OOB data (#423) on a per-host basis.
Some of the configs such as `hostname`, `agent forwarding`, `X11 forwarding`, etc... could actually be extracted from .ssh/config directly.

I think it's much better than aliases because it's seamless and consistent. Plus the fact that mosh already seamlessly use .ssh/config makes it quite irritating to be able to use aliases only for specifying a couple of argument for a given host. If many hosts requires a custom config multiplication of aliases could also clutter the shell. Plus we always have to remember which hosts works out-of-the-box and which needs an alias. All this disappears with a `.mosh/config` config file.

Security-wise, since it's already done by ssh (and actually allows a much bigger control on the connection since it could change the Hostname we are connecting to, define a ProxyCommand, etc...), If the file is secure, I don't see a very big security issue here.

infinity0:
Instead of having a separate `.moshrc`, it would be easier for users, to just put these extra options in `.ssh/config`?

~~~~
Host xxx.com.yyy
HostName xxx.com
User uuu
MoshPort 6nnn
MoshPrediction always

andersk:
@infinity0, nope, the unrecognized syntax would completely break `ssh` (and therefore also `mosh`). I could imagine putting them inside magic comments…but not for two seconds longer than I have to.

emtiu:
Any consensus on this in sight? I'm also very much interested in server-side custom port(range)s. Don't mind if it's systemwide (as in /etc/moshports) or per-user (as in ~/.moshports).

Toub:
Also interested to have a client-side or server-side configuration for udp port range.

rgruesbeck:
+1 for this

Myke:
Another vote for server-side config; with multiple servers behind NAT, ssh-config nicely solves the port changes, but there's no elegant way to teach mosh that it's only getting specific ports forwarded to it.

yookoala:
+1

[...]

PR #501, opened 2014-02-09
Issues/431/dev: configuration file implementation

earlchew:


cgull:
I'm a little uncomfortable with adding an ad-hoc config file parser in a security-sensitive program, and I'm not sure SSH is a good example to follow, perhaps YAML might be better here.

I echo @andersk in saying surely there must be a simpler way, though I don't know what it is either.

glensc:
testing `ConfigFile` package inside main mosh script? does not seem like super idea.

maybe use existing CPAN module to parse the config, and if want to avoid depending on CPAN modules in distribution (why?), could use some other module that inlines needed CPAN modules during build process. Can't give any reference which tool it is, as haven't used any myself, but there exists, maybe even several projects.

ekacnet:
FYI my PR. is similar to this but:
I'm only adding ~130 lines, most of commits are cosmetics and are meant for reducing the risk of forgetting options (ie. by not having them spread all across the script).

It's really handy to have the capacity to set some options like predict or the path to the server command, it's also handy if you have to use a custom ssh command as I do when I do port knocking or use 2FA like yubikeys

achernya:
Thanks for this change. However, I think it takes an approach that is different from how we've decided we want to go in #431, and currently has a lot of conflicts for merging, so closing in its current form.

PR #509, opened 2014-04-15
Implement CTTY grabbing when TIOCSCTTY is missing

glance-:
AIX is missing TIOCSCTTY, so we assign us a ctty via a open of the slave
device.

Signed-off-by: Anton Lundin <glance@acc.umu.se>


cgull:
Pulled.

PR #511, opened 2014-04-18
Timestamp: Prevent integer overflow on Darwin PPC 32-bit

### LINKED ISSUE #479: mosh doesn't work on PowerPC OS X 10.5 (Leopard)
Built 1.2.4 from source after trying Tigerbrew's 1.2.4.  Both seem to connect and then instantly exit, saying

    [mosh is exiting.]

After a couple of line feeds.  Tried on a couple different mosh servers, x86 and ARM, both Debian or close enough.

andersk:
A Darwin PPC 32-bit user observes huge values `numer == 1000000000` and `denom == 18431683` returned from `mach_timebase_info()`.  For these values, `mach_absolute_time() * numer` overflows `uint64_t` every 1000.82 seconds, and `1000000 * denom` always overflows `uint32_t`, with the effect of making time run backwards at -11190660 times its usual speed.

This bug was masked on Darwin x86 64-bit, where `numer == denom == 1`.

Fix it by doing the conversion with `double` arithmetic instead.

Closes #479.

### DIFF
--- a/src/util/timestamp.cc
+++ b/src/util/timestamp.cc
@@ -72,14 +72,16 @@ void freeze_timestamp( void )
   }
 #elif HAVE_MACH_ABSOLUTE_TIME
   static mach_timebase_info_data_t s_timebase_info;
+  static double absolute_to_millis;
 
   if (s_timebase_info.denom == 0) {
     mach_timebase_info(&s_timebase_info);
+    absolute_to_millis = 1e-6 * s_timebase_info.numer / s_timebase_info.denom;
   }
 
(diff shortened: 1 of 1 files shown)


cgull:
Have any of you tried this on Darwin/PPC?  Once it's tested there I'll pull it.

mistydemeo:
Yep, both @gordon-morehouse and I have tested this and it works well.

cgull:
Pulled.

gordon-morehouse:
Hand-me-down users the world over thank you! ;)

PR #514, opened 2014-04-26
Conditionally rename main to mosh_main for building under NaCl.

rpwoodbu:
This is the one and only patch I have in Mosh for Chrome to be able to use Mosh as a library in my NaCl wrapper. It'd be great to have it upstream so I don't have to worry about it anymore.


cgull:
@rpwoodbu: don't pull that yet, I may rebase `keithw:master` to get rid of the merge commits.

andersk:
@cgull: No. Never rebase published the published history of `master`.

PR #518, opened 2014-05-05
bash completion: Support ~/.ssh/config aliases

andersk:
`_known_hosts` does not complete aliases defined in ~/.ssh/config by default, but it takes an `-a` option to do so.


jtokoph:
:+1:

cgull:
Incorporated into #628 / #642.

PR #522, opened 2014-05-18
Add <esc>-^L for local repaint, fix minor escape-key bugs

### LINKED ISSUE #370: A way to trigger a mosh-client repaint without sending ^L to the remote process
Mosh client got into a slightly inconsistent state and I wanted the local process to repaint the screen. The way to do that was to hit ^L, which also sends ^L to the remote process. In this case my remote process responds to ^L not with a simple repaint but by clearing the screen, and that was undesirable.

It would be nice if some keybinding like ^^^L would do a local repaint and not send anything to the remote process.

cgull:
This adds <escape>-^L as a way to repaint the client's screen without involving the server.  It also makes prediction work more correctly on escape-key sequences, not that it matters a whole lot.  This resolves #370.

Additionally, I think the repaint-on-normal-^L is kind of a bad idea, but I haven't done anything about that here.

### DIFF
--- a/src/frontend/stmclient.cc
+++ b/src/frontend/stmclient.cc
@@ -319,8 +319,6 @@ bool STMClient::process_user_input( int fd )
     for ( int i = 0; i < bytes_read; i++ ) {
       char the_byte = buf[ i ];
 
-      overlays.get_prediction_engine().new_user_byte( the_byte, *local_framebuffer );
-
       if ( quit_sequence_started ) {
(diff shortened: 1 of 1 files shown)


eminence:
This is old.  Seems like we should accept this, or close the PR.

yury:
I don't is it common case in general. But vim users inside tmux map navigation between panes with CTRL+hjkl

and CTRL+L triggers ugly blinking...

https://youtu.be/LcaijV7Nj44

PR #523, opened 2014-05-18
allow/prefer a system OpenSSL

cgull:
FreeBSD (and the other BSDs) bundle OpenSSL into the base system.  They may not have it installed as a package.  Attempting to configure Mosh on such a system will fail.  This builds fine on the Linux systems I have available but it might go wrong someplace else.

There are other portability fixes in the performance branch which I will be pushing shortly, I didn't fully disentangle all my commits.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -243,7 +243,8 @@ AC_CHECK_FUNCS([m4_normalize([
 
 AC_SEARCH_LIBS([clock_gettime], [rt], [AC_DEFINE([HAVE_CLOCK_GETTIME], [1], [Define if clock_gettime is available.])])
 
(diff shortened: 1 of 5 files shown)


cgull:
Closing in favor of #610, which is just the non-OpenSSL portability fixes in here (plus one more I missed).  The OpenSSL issue has an easy workaround, the autoconf hackery was a bit fragile, and the *BSD bundled SSL libraries are moving targets just now (LibreSSL, and there is FreeBSD talk of making their bundled OpenSSL private).

cgull:
I came up with a better solution in #686-- support other crypto libraries and avoid OpenSSL hackery altogether.

PR #530, opened 2014-06-03
bind socket using a source port from MOSH_SPORT environment variable

meeuw:
Hi Keith,

Could you please send me some feedback about this crude hack?

It seems to work and I'd like to see it included in mosh.

Thanks in advance!
Dick Marinus

### DIFF
--- a/src/network/network.cc
+++ b/src/network/network.cc
@@ -396,6 +396,17 @@ Connection::Connection( const char *key_str, const char *ip, const char *port ) 
   socks.push_back( Socket( remote_addr.sa.sa_family ) );
 
   set_MTU( remote_addr.sa.sa_family );
+  Socket socket = Socket( remote_addr.sa.sa_family );
+  struct sockaddr_in sin;
+  memset(&sin, 0, sizeof sin);
+  sin.sin_family = AF_INET;
+  sin.sin_addr.s_addr = htonl(INADDR_ANY);
+  int sport = 0;
+  char *env_sport = getenv( "MOSH_SPORT" );
+  if (env_sport != NULL) sport = atoi(env_sport);
+  sin.sin_port = htons(sport);
+  bind(socket.fd(), (struct sockaddr *)&sin, sizeof(sin));
+  socks.push_back(socket);
 }
 
 void Connection::send( string s )


flixr:
I have a very unusual use-case, but would benefit from this as well.
:+1: from me

eminence:
Bumping this issue.  I maintain this same functionality with an out-of-tree patch and I'd love to be able to use a stock mosh

xtaran:
Please don't name that variable `MOSH_SPORT`. I really wondered what kind of sport is involved with Mosh. :-)

flixr:
FYI, I additionally disabled client port hopping if MOSH_SPORT is set: ed0d32d72b855527167ba7ca08b47476f7403687

eminence:
FWIW, here is my version of this patch:  https://github.com/eminence/mosh/commit/7e0c3eef1105d326af88b8ca2a2b8aaef5a17d20

meeuw:
@xtaran I've used SPORT from the same parameter as iptables uses (iptables -p tcp --sport)

As @keithw asked, what is YOUR use case for this patch? Mine was only a proof of concept for connecting through a NAT from the Internet, which failed miserably because NAT timed the connection out and the (translated) port changed.

I've never encountered a firewall which allowed only a specific udp sport/dport tuple, what would be the purpose for this?

The only real use case I could think of is (I'm not sure if it would work) that you can kill mosh-client and restart it with the key+sport and be connected again.

eminence:
My use case is to allow a "reverse connection".  The scenario is this:

mosh-server is behind a NAT/Firewall that I cannot change.  The server machine does not accept SSH connections from the internet, but I can SSH to it if I am on its LAN.  The NAT/Firewall doesn't allow unsolicited incoming UDP packets, but once a "connection" has been established (by sending an outgoing UDP packet), return (incoming) packets are allowed.

mosh-client is on an internet-facing device, with no address translation, and no firewall

Given the following network diagram, here is an example:

```
  _____________                           _____________
 [ mosh-server ]    [N]                  [ mosh-client ]
 [ 0.0.0.0:SP1 ]====[A]====[internet]====[ 0.0.0.0:CP1 ]
 [_____________]    [T]                  [_____________]

```

A process on the `mosh-server` machine binds to local port `SP1` and sends a UDP packet to `(mosh-client, CP1)`.  A process on the `mosh-client` machine sees this incoming packet as coming from `(nat, SP2)`, and sends a reply to that address (knowing that the NAT has mapped `SP1` to `SP2`.  The `mosh-server` machine receives this packet.

Now that a set of round-trip packets have been successfully sent/received, the `mosh-server` process starts up, and binds to local port `SP1` (this is easily done via the command line).   The `MOSH_KEY` is read and communicated OOB to the `mosh-client`.  `mosh-client` starts up, and binds to `CP1` (via an environment variable from this patch), plus the `(nat, SP2)`.  

Now mosh just does its normal thing -- the client and server can communicate directly.

I have some wrapper scripts that handle this packet sending stuff, and takes care of communicating the key back to the client machine, but I have to have a custom version of mosh-client to make this work.

Issue #372, opened 2013-01-14
mosh-server on Raspberry Pi: ~50% cpu load


phmarek:
See also #297; "top" with 1 sec interval causes mosh-server to eat nearly half of the CPU.

"-pg" profile available.

vielmetti:
Ran across this issue and wanted to report current (March 2015) timings:

```

  PID USER      PR  NI  VIRT  RES  SHR S %CPU %MEM     TIME+ COMMAND                        
 5392 pi        20   0 14880 3448 1280 S  7.2  0.8   0:12.69 mosh-server        
```

7.2% of CPU seems a little high but is by no means outlandish, this is running top on a Pi.

gordon-morehouse:
I'm seeing 5-18% with a steady state around 12.5% on my Pi, which isn't overclocked to the max (I think I have it at 900MHz core clock, not 1GHz).

This is mosh 1.2.3-1 on Raspbian; I've had no noticeable problems with mosh-server and CPU usage on a Pi supporting a few users.

cgull:
It occurs to me that Raspbian has probably seen performance improvements since 2012, and that might explain the different results people are getting.

That said, my changes in #531 will get you about an order of magnitude better performance.

cgull:
Closing, please open a new issue if you still have performance issues.

PR #531, opened 2014-06-15
  Performance improvements for Terminal::Framebuffer and Terminal::Display

cgull:
This replaces pull request #527.  This pull request has had some extraneous stuff weeded out, some pushed to other branches, bugs fixed, and (most important) a bunch of time spent in 'git merge -i' to its benefit.  But the discussion on that pull request is still background for this one.  I've fixed all the bugs I know of, the pull request is comprehensible now, and I think it's ready for Mosh.

To summarize what this does:

* It improves the containers in Terminal::Framebuffer.  Terminal::Renditions and Terminal::Cells go on a diet (important when you have 30k of them), Cells use UTF-8 strings instead of vector<wchar_t> because std::string is faster everywhere I've tried it (even with different implementation strategies), and Rows are stored in vectors instead of deques (deques do neatly handle scrolling, but they are costly to index and construct/delete).  It also reworks some methods to allow slightly more efficient generation of the diff string.
* It rips apart^W^Wrefactors Terminal::Display::new_frame() and put_cell() to use a put_row() instead, and carefully optimizes the per-character loop inside to be much more efficient, and more logically organizes the state necessary for tracking the generation of the diff string.  It also reworks character output to reduce the amount of characters generated significantly.  Another motivating piece for this was to make both Framebuffers passed into new_frame() const references, instead of having to copy one.  That in itself improved performance, but it lead to...
* The last and largest improvement: making the Rows in Framebuffers shared, copy-on-write objects with the help of std::shared_ptr.  I knew something like this was possible, but scratched my head for a while on how to do it cleanly.  The change ends up being elegant and not very large, though-- thanks to past efforts to ensure const correctness everywhere else in Mosh.  This greatly reduces the amount of copying mosh has to do to maintain multiple terminal states (except for one case-- where you cat a large file to your terminal and it's scrolling as fast as it can).  This greatly improves Mosh's performance for actual use cases where not much of the display changes from frame to frame, like sitting idle or working in an editor.

It also has various little algorithmic improvements and fixes for pet peeves, but those are fairly straightforward stuff.

I think the shared/COW Rows change also has potential to help with other long-standing feature requests and issues in Mosh, such as versioned prediction and scrollback buffering.  I haven't really explored that though.


gordon-morehouse:
If the claim of far better performance on Raspberry Pi is true (see #372), it'd be nice to get this merged for those of us using mosh daily on the Pi and other machines with very little horsepower like routers, etc commonly used as gateway hosts.

cgull:
In commenting on #372, I tossed off a 10x improvement quip.  It is, of course, more complex than that.  10x is a rough approximation for what I saw on 300x100 windows in development.

mosh's CPU consumption is mostly related to the size of the terminal window.  My changes reduce the algorithmic complexity of frame comparison/redraw, so the bigger your window gets, the larger your relative improvement gets.  The type of output has an effect as well-- my changes have the most effect when there are small changes on a mostly static screen, such as when you're typing in an editor or at a prompt.  An invocation of top is improved less, because it does a single repaint of the screen, possibly changing a lot of text, every 2 seconds.

So depending on what you're doing the improvement might be as little as 1.5-2x and as much as 100x.  At this point, I don't see the new code slowing things down anywhere, though I'm sure pathological cases can still be found.  80x65 is about the smallest terminal window I'll use for real work (you can have 2-3 of those on a 1366x768 display), the off-the-cuff factor for those is still 3-4x improvement.

Another way of looking at the performance improvements is this:  Before, Mosh was far and away the largest consumer of CPU for interactive sessions.  Now, the Emacs you run in that session often consumes more CPU managing its own screen redraw.  I didn't look closely, but I recall some subset of tmux, vim, and Emacs showing scaling issues, and BSD vi doing better with larger windows.  Before, Mosh was an unusable slide show on an RPi with large windows.

gordon-morehouse:
Are there clear security implications here or is this something that could be merged without a fine-tooth-comb audit?

cgull:
These changes don't interact with mosh's crypto or network code at all.  
So no direct security implications, just the usual issues of new code 
that might have exploitable bugs.  It's been a while, but I think I did 
some review with an eye on security, and I ran this code through static 
analysis with clang and Coverity.  The last Coverity run is a bit stale, 
time for me to update it.

vielmetti:
I'm interested in trying this out on my Raspberry Pi (running Raspbian). Do you have any simple build hints so that I get exactly the right version to test?

cgull:
That shouldn't be any more complicated than:
```shell
git clone https://cgull/mosh.git
cd mosh
git checkout performance2
./autogen.sh
./configure
make
make install
```
which is the standard procedure for building Mosh from github, except that you're getting different source code.

vielmetti:
Aha, building on Mac OS X ran into

https://github.com/keithw/mosh/issues/251

which is "don't have all the dependencies installed", namely `protobuf`. Digging some more.

vielmetti:
After I installed `protobuf`, it built. What's the best way to test (there's not a `make test` target).

vielmetti:
On my Mac I was able to successfully make a connection with this new mosh:

```
edwards-air:scripts emv$ pwd
/Users/emv/src/mosh/scripts
edwards-air:scripts emv$    ./mosh --client=../src/frontend/mosh-client emv@10.7.0.1
[mosh is exiting.]
```

I *think* that gets the test bits right. Let me spin up a machine to put the server on too.

cgull:
The performance improvements are more significant on the server side, the client generally has to do less work.
For synthetic performance testing, there's the stuff in `src/examples`.  `termemu` is the one most applicable to this work.

vielmetti:
Now building on a Ubuntu VM and it doesn't have autoreconf

```
emv@vagrant-ubuntu-trusty-64:~/src/mosh$ ./autogen.sh
./autogen.sh: 3: exec: autoreconf: not found
```

which means I need to pull in the `autoconf` package. Doing that now.

gordon-morehouse:
Thanks for the hard work on this, sooner or later a lot of overworked Raspberry Pis will breathe easier! ;)

cgull:
For the record:  my branch doesn't change Mosh's build-time dependencies significantly, so you can just use your system's package system to get build dependencies and then build my code.

bdrewery:
Can this be merged for 1.2.6?

cgull:
I was a little overcommitted on the 1.2.5 release, plus summer, plus wanting to add a test framework for Mosh.  It'll happen, I just need to figure out priorities and ordering.

ctrlcctrlv:
Awesome :)

cgull:
I've added the test framework, and I've added some tests relevant to this code (and a couple of bugfixes for things found that way).  Every commit passes 'make distcheck' now, too-- there were a couple of things that had ended up in the wrong commit in previous rebasing.  Final review here, folks, I will pull soon if I don't get any thumbs down or other worthwhile review.

keithw:
Looks really good. I tried to give it a torture test and it passed everything, with one exception. We no longer have a limit on the number of combining characters that can go in a character cell, which seems to produce two badnesses if you run something evil like `perl -we 'print "a"; while (1) { print "\314\202" }'`

(1) `mosh-server` starts to use lots of RAM, and the outer terminal emulator and `mosh-client` can get totally hosed.
(2) If `mosh-server` was started with `new -v`, it will start to log roundtrip failure warnings!

#1 is not a big surprise and can probably be fixed by reinstating the limit on the number of combining characters in a cell, but are you able to reproduce #2 and (if so) figure out where the roundtrip failures are coming from?

cgull:
I restored the combining characters limit in my working sources, coming soon to Github.

The roundtrip failures have nothing to do with my changes, I have test cases that can provoke them on Mosh master.  I think I've found the issue-- the server terminal state may be copied while UTF8Parser has bytes buffered, so you need to clear that before you try applying diffs to it for whatever reason.  I'm not sure why this hasn't caused more trouble.  This seems to fix it: https://github.com/cgull/mosh/commit/7bacfecdd9fb9df74b178d57437937b409eab593  The same might also apply to ANSI X3.64 parser state.

cgull:
Here's the (messy) reproduction for the roundtrip failures, it just fills the screen with `U+C111`:
`sleep 3; clear; perl -we 'for my $h (1..37960) { print "\xec\x84\x91"; for my $i (1..1) { print "\314\202" if 0 }}; print "\n"; sleep 2'`
You need a big, big window for the problem to show clearly.  The one I've got now is 520x149, and it shows very well.

keithw:
Having trouble reproing this. :-(

cgull:
The original reproduction was on FreeBSD 10.2 on an 8-core desktop.  But I just reproduced this first try on an old single-core laptop running Ubuntu 14.04.

I wish *I* were as smart as I was in 2011.  So I'll give you the longer explanation and maybe I'll get it in return someday...

* tty/termios is a stream interface that applications write to mosh-server on.  At any time, but certainly when mosh is busy and not keeping up with termio and tty buffers are full, mosh-server may read bytes in arbitrary segments having nothing to do with the segments that the application wrote.  In particular, the segmentation might split UTF-8 characters or ANSI escape sequences.
* mosh-server applies the bytes to the running terminal state.  That might leave UTF8Parser waiting with `buflen` bytes in `buffer`, or Parser in something other than Ground state.
* mosh-server then copies the running terminal buffer into current_state, which is the start of state sync management.  This includes Parser and UTF8Parser state, of course.
* mosh-server generates state diffs to send to mosh-client.  These include a character stream that is a well-formed sequence of UTF-8 and ANSI sequences; they are always complete.
* For the round-trip verification, mosh-server copies a previous state and applies a diff to it, then compares the now-modified framebuffer with current_state's framebuffer.
* **fail**:  When applying that diff, the UTF8Parser or Parser state may be expecting the remnants of whatever they were expecting.  And so you get corrupt UTF-8 and corrupt printing input, and could also get corrupt ANSI sequences causing who knows what to happen on that frame buffer.  I think this *only* happens on mosh-server, and *only* in the round-trip verification code.  mosh-server never has to apply diffs otherwise, and mosh-client applies diffs all the time-- but they are always the well-formed ones from mosh-server, never raw tty input.  The user never sees any problem, because there *isn't* a problem on the code/data flow for normal Mosh operation.
* **fix**:  For round-trip verification, clear the UTF8Parser and Parser states to their ground/initial states.  Then, applying well-formed diffs will not cause these hiccups.  (My diff above only clears UTF8Parser and has a bunch of other random debug stuff in it.  But that's enough to fix the repro.)

cgull:
Also reproduces first try on OS X 10.10 on an iMac Retina with iTerm 2.  The command line in all of these is `mosh-keithw --server=bin/mksl --local 127.0.0.1 2> /tmp/client.log`, where `mosh-keithw` is a wrapper script that invokes Mosh with the paths for my Mosh working directory:
```shell
#!/bin/sh
MOSHNAME=$(basename $0)
if [ $MOSHNAME = mosh-test ]; then
  MOSHNAME=mosh
fi
~/src/$MOSHNAME/scripts/mosh --client=$HOME/src/$MOSHNAME/src/frontend/mosh-client --server=src/$MOSHNAME/src/frontend/mosh-server "$@"
```
and `mksl` is
```shell
#!/bin/sh
~/src/mosh-keithw/src/frontend/mosh-server "$@" -v 2> /tmp/mosh-server.$$.log
```

keithw:
Aha! It hadn't occurred to me that the bug only affects the round-trip verification and not the normal use of Mosh. Thanks for all the detail; I'm sure us-of-2019 will be grateful for it!

cgull:
Do you have any ideas on how to make the round-trip verification a stronger/more useful test, now that it's working right?  It's potentially really valuable for finding bugs in diff generation, of course.

I think I will enable/check it in the new test framework but if we can find a way to apply it and report errors in interactive usage in the wild, that'd be great.

cgull:
@keithw:  would you look at my WIP on https://github.com/cgull/mosh/tree/roundtrip-fix-2 (specifically https://github.com/cgull/mosh/commit/2048f17904dd95610be75fbb04fb85c5af1b7278 and https://github.com/cgull/mosh/commit/9af3603dffa5393af27c378ab7becd8e25546a0f) to see if those account for the round-trip warnings you were getting?

That branch fixes the round-trip warnings I was seeing, but my latest fix has caused a couple more to crop up-- but only from a new, extended Instruction verification I added.

cgull:
I've rebased that WIP into final product.  The new errors I was seeing appear to be testing artifacts caused by variation between using spaces or erase commands to clear rows, and don't seem to indicate real problems; I decided to ignore the new verification check that produces them for now until we have some more experience with it.  If you'd check this against your torture test, that'd be great.

keithw:
Ran this through some tests. One issue: with current Mosh master, if an application prints bytes to the screen that don't change the terminal state, nothing even gets sent to the client. With this, the `mosh-server` sends a continuous stream of updates to the client, even when nothing (?) is changing on the screen.

Example: `while true; do echo -ne "\033[K"; done`

Afraid I'm not sure if this is an issue with the Terminal::Complete `operator==` or what.

cgull:
Interesting catch.  It's because `Terminal::Framebuffer::operator==()` is now shallower than it used to be.

* `CSI_EL` writes blanks to the end of the row.  This does COW on the row, and of course changes its pointer in `Terminal::Framebuffer::rows`.
* `Terminal::Framebuffer::operator==()` compares the `rows`, gets a difference, and that causes `new_frame()` to get called.

The solution for your specific example is easy enough-- just do a deeper compare when the pointers differ-- but I'm a little concerned about other possible cases, such as '\033[J', which would make that more expensive.  There's also the possibility of cases where we're writing spaces into cells and we'd like them to compare as equal.  I'll think about this a bit.  Do you have other specific cases where you saw this behavior?

keithw:
I believe I also saw it with the earlier case of just adding more and more combining characters.

cgull:
I'm good.  I'm so good I fixed this issue before either of us knew about it.  :)

I'd written some code in an attempt to improve the visual mess you get when you resize a tmux window.  It doesn't seem to help much for that but it also drops empty frame diff strings, which means that state sync diff strings can be empty, which means that we can avoid sending them at all.  I think that's a better approach for now, it avoids unexpected changes to terminal rendering or the performance issues I worked on.

As for the combining-chars case:  This doesn't fix that, because the grapheme is genuinely (if not exactly usefully) changed.  I think we have to get a *lot* smarter about combining marks to do anything useful there.

keithw:
Just rebased this on master and ran it through some tests. All lgtm. (In my view, the case of an app that prints tons and tons of combining characters to the same cell is now properly handled -- after the limit on cell size is reached, the server-side framebuffer stops changing, and the `mosh-server` stops sending any updates to the `mosh-client`).

cgull:
Now in master.  If anyone sees display corruption, cut/paste issues, or anything else, please report it.

wincent:
Congrats, @cgull. This PR is an epic achievement.

mbsulliv:
FYI, I just built with 90a529b today in a windows cygwin environment and one of the changes between 1.2.5 and 90a529b (I'm guessing this one) fixed a slowdown I was having in vim with multiple windows where scrolling via 10CTRL+Y or 10CTRL+E was very slow to repaint.  It always worked fine in linux, but not in cygwin.  Now it's working in cygwin :)

cgull:
Closing, I've committed code that should fix this.  Please open a new issue if you still have performance issues.

hryamzik:
Is it available on released versions or only on dev? I'm facing hight CPU usage on Raspberry Pi with ubuntu 16.04.

cgull:
@hryamzik, this work is in Mosh 1.2.6.  If you're using Raspbian, the package there is still 1.2.4a, and the server will be slow, especially with large windows.

If you can install a newer version, you will see a large improvement.  I run my IRC client on a Pi, and the poor performance there was the primary motivation for fixing it.

hryamzik:
Thanks @cgull! Looks like mosh 1.2.6 wasn't released as PPA, I wonder what's the showstopper...

cgull:
Since Raspbian is an unofficial Debian port/fork, we can't add it to the main Debian Mosh PPA; I think we'd have to maintain a manually-built PPA for Raspbian.  I'll open an issue for this idea.

hryamzik:
@cgull I'm talking about ubuntu, there's a [dev PPA](https://launchpad.net/~keithw/+archive/ubuntu/mosh-dev) for it but no stable one.

PR #553, opened 2014-09-29
scripts/mosh: Allow IO::Socket::INET6 as an alternative to IO::Socket::IP

andersk:
Some systems have IO::Socket::INET6 preinstalled, so we might as well accept either.


cgull:
Pulled.  @keithw, I'm satisfied by both https://wikispaces.psu.edu/display/ipv6/IPv6+on+OS+X and inspection of an old OS X 10.5 install that this will make IPv6 work for Mac users going back that far.  I also added a better error message than the opaque one Perl spits out-- but if anyone has suggestions for better wording, do tell.

PR #564, opened 2014-11-02
Add support for configuration file and 2FA like yubikey

ekacnet:
6 changes most of them are clean up to support properly the configuration files.

We have configuration file like that:

Host *
   ssh = ssh
Host poseidon  
    ssh = ssh -p 22                                                             

host scl64r1_1  
    port = 2239  
    predict = never

Last change add 2FA with yubikey/duosecurity

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -62,6 +62,10 @@ my $bind_ip = undef;
 
 my $port_request = undef;
 
+my $no_redirect = 0;
+
+my @ssh_options = ('-S', 'none', '-n', '-tt');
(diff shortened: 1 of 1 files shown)


ekacnet:
Any news on this PR?

cgull:
This is very similar to #501 and has similar issues. There's no consensus yet on whether a config file should be supported and how it should work.

achernya:
Thanks for this change. Unfortunately this goes in a different direction than we've decided to take in #431, and now has merge conflicts, so closing in its current form.

Issue #560, opened 2014-10-11
Mouse reporting stops on window resize


wlcx:
With a mosh session in iTerm2 on OSX and tmux, mouse reporting works and the cursor is displayed with a circle on the bar. Resizing the window in any direction stops mouse reporting from working, with the cursor changing to a non-circled state. Using ssh, mouse reporting stays active on window resize.

cgull:
I think this is resolved by #576, it tests out OK for me.  See that issue for a possible explanation of what causes it.

Issue #505, opened 2014-03-19
Mouse support does not work 100%


creack:
#461 is awesome! It enabled the mouse support.

However, this doesn't work at 100%. Emacs has difficulities catching the mouse (I have to disable/reenable the xterm-mouse-mode) and when I try to select something it tells me `Unexpected sequence from XTerm`. It works fine on SSH.

The selection works fine on tmux though.

```
$> emacs --version
emacsclient 24.3
```

dbkaplun:
:+1:

cgull:
I *think* this is fixed by #569 and/or #576.  In any case, the code has changed, and any issues should be reported anew.

PR #461, opened 2013-09-09
Implement xterm mouse modes

niieani:

PR #576, opened 2014-12-29
Fix xterm mouse support.

lpkruger:


sunlight07:
I have tested this and it seems to work very well. @keithw , could this be merged into your master branch?

sunlight07:
@keithw, could this be merged into master? I have tested it and it seems to work very well.

Pan0ram1x:
Confirmed. Please merge this.

Kriechi:
Yep - works for me! Thanks!
:+1:

jwon:
I confirm that this works great! :+1: Please merge.

jwon:
So this pull request works great for mouse events, but I still have a weird bug.

So I'm trying to use mosh+tmux+vim. I have I have in my .vimrc:
```
set mouse=a
```
and in my .tmux.conf:
```
set -g mode-mouse on
set -g mouse-select-pane on
set -g mouse-resize-pane on
set -g mouse-select-window on
```

First of all, I would like to say that any time I am in vim, mouse events work perfectly and everything is dandy. HOWEVER, any time I change my focus to another tmux window/pane that DOESN'T have vim open (let's say just a normal bash prompt), as soon as I change to that window/pane I lose mouse support. The only way for me to get mouse-events back is to change my focus to a window/pane that has vim open (using keyboard hotkeys), and then mouse-events will come back. 

Any ideas? :(

dwf:
Works for me! Homebrew formula for the interested here: https://gist.github.com/dwf/26a13e03009068b4299a

Hope this gets merged soon.

jagtesh:
It works, please merge this!

cgull:
@lpkruger: I'd like to merge this, but it's, erm, underdocumented.  What does it fix?  What does it now support?  Where has it been tested?

dwf:
@cgull As of current master, if you, say, run tmux within mosh with mouse
mode enabled, mouse event for generating a selection works precisely once,
after which the mouse event is no longer is captured by tmux and is handled
by your terminal emulator (I've only tested this on iTerm2 on OS X, but I
imagine this is reproducible on other terminals; perhaps another one of the
commenters can confirm this). With @lpkruger's patch it's working the same
as when I'm logged in via SSH.

cgull:
There's a number of issues here.
- what does the patch *do*, exactly?  (I've since found that XTerm's documentation specifies mouse support pretty well.  There are around 4 different modes.)
- What console applications have been tested?
- What terminals have been tested?

There's a bit of an M*N flavor to this, unfortunately, and I was hoping @lpkruger would tell us what he knows.  And I'm not a great tester for this, I don't like or use mouse functionality on console apps much.

cgull:
This appears to be a reasonable subset of what's in `xterm`-- http://invisible-island.net/xterm/ctlseqs/ctlseqs.html
Unsupported:
- alternate-screen mouse-wheel mode isn't supported (but we haven't got alternate screens yet).
- focus events aren't supported.
- highlight mode won't actually work; it needs the sequence `CSI Ps ; Ps ; Ps ; Ps ; Ps T` to be implemented.  I doubt anybody cares, this looks like an experimental hack from the early days of mice.

cgull:
This issue is a followon to #461, which has lots of relevant links to useful things.

cgull:
Pulled, because this is clearly a big improvement on what we had before.

I see a possible issue though:  If a terminal emulator doesn't support encoding mode 1006 (SGR), and an application requests, say, modes 1005 and 1006, then we may have a problem.  This patch contains the encoding mode as a single state.  Depending on the timing of reception of the application's requests, Mosh may or may not send 1005 with one frame update, but will send mode 1006 with another frame update.  In this case, the terminal emulator will stay in default encoding mode, which reports incorrectly past column or row 223.  I think this issue also applies similarly to the reporting mode.

In practice though, it all seems to work as well as it can-- the remaining issue is that terminal emulators that don't support encoding mode 1006 (urxvt, for example) break when you point with the mouse past column/row 223, with any combination of applications, screen/tmux, mosh/ssh.

In my testing, the only terminal emulators that have support for encoding mode 1006 are
* iTerm2
* Putty
* Cygwin newterm (not surprising, it's derived from Putty)
* XTerm
I strongly suggest these if you like having large windows open.  Also, it appears that vim does not support encoding mode 1005/1006, and always breaks past column 223.

I think for a truly pedantically correct fix, Mosh would have to record the individual mouse reporting/encoding codes sent by the application, the individual states they set, and *the order in which they were received*, and replay them to the terminal *in that same order* (including when restarted after `^^ ^Z`).
But I would guess that the applications that try multiple modes shotgun the terminal with numerically ascending sequences, and just take whatever bytes they get back from the terminal, so in practice I think this fix will work OK most of the time.  I hope.

### Testing notes:
I tested a subset of @eminence's scenarios in #505 (basically all of his scenarios with XTerm, and some with urxvt).  I also tested:

apps:
* vim: click, mouse-wheel scroll
* emacs: click, drag, mouse-wheel scroll
* weechat: click, mouse-wheel scroll in user list
* elinks: (achin) click, mouse-wheel scroll
* htop: click

test programs:
* xterm/vttest
* http://www.leonerd.org.uk/code/libtermkey/
* mousetest.pl https://code.google.com/p/iterm2/issues/detail?id=432
 (now at https://gitlab.com/gnachman/iterm2/issues/432, *very* smooth
 move on issues)

terminal emulators:
* Terminal (no XTerm mouse support)
* iTerm 2 (full support)
* xterm (full support)
* Putty (full support)
* Cygwin (full support)
* gnome terminal (no SGR)
* Konsole (no SGR)
* xfce terminal (no SGR)
* urxvt (no SGR)

virtual terminals?
* tmux: enable mouse mode, click to select windows
* screen: nothing much?

particular issues:
* wide terminals 
* ^Z in Mosh (bug found, fixed)
* window resize #560 (works now)

Testing mouse issues is a combinatorial explosion, it would take a week or two from a full-time QA person to work through all the possible cases.

I'm closing all the other open mouse issues and redirecting comment here.

Issue #12, opened 2012-02-14
Reconnection and garbage collection


keithw:
Allow reconnection to existing sessions that were somehow severed. Give command to clean up disconnected sessions.

royaltm:
kill -1 works as well, no need to -9

cgull:
@spcomb has a good point there...but I have another idea for garbage collection that may help individual users.

`mosh-client` could save a file with a prepared shutdown message, with a high sequence number (2^48 - 10 sounds about right).  This is encrypted and ready to go out on the wire.  It holds an fd open on the file, and takes a write lock on it.  It unlinks the file on normal exit.

At subsequent `mosh` startups (or some other convenient time), `mosh` (or a future `mosh-cleanup`) could go looking for these files.  For each that it can open and get a write lock on, it sends the message and unlinks the file.  Perhaps it only sends messages to the same server it's currently connecting to, not sure yet.

I don't see any obvious security issues, other than that anybody who gets hold of that file can send the message and DoS the session.  The adds to mosh-client aren't trivial, but they're not large either.  But I've not thought about it too hard yet.

This doesn't work on the server, because clients are mobile, and we wouldn't know where to send the shutdown message.

cgull:
Some IRC chatter:

2015-10-16 14:19:11     achin   $ ps -ef |grep -i mosh-server |wc -l
2015-10-16 14:19:14     achin   54
2015-10-16 14:25:48     KeithW  I think we want a feature (stealing from rex) where every *new* mosh-server checks for dormant mosh-servers that belong to mosh-clients on the "same machine" as the current mosh-client, and if they're currently dormant and there's no mosh-client still alive on the "same client," it kills them.
2015-10-16 14:29:59     achin   it would be up to the client to include a reasonable clientID when connecting?
2015-10-16 14:30:19     ddrown  I like this client ID idea
2015-10-16 14:33:39     luto    stating the obvious: the new mosh-server must only look for other mosh-servers running as the same server-side unix-user
2015-10-16 14:36:45     KeithW  achin: Yeah. And then it would be up to new clients to report all the "still-running mosh-clients on the same ClientID machine" to the new mosh-server, so it can call all the mosh-servers (from that ClientID) that are no longer claimed.
2015-10-16 14:37:00     KeithW  But figuring out how to do this robustly takes some work...
2015-10-16 14:39:33     achin   i like this idea
[...]

PR #580, opened 2015-01-20
Prevent defunct mosh-server processes from piling up

palmer-dabbelt:
I've recently started using Mosh, and it seems really useful.  I was wondering what happens when my most-client dies, and according to some GitHub issues it appears that mosh-server just keeps running indefinately.  This seems bad to me, so I went ahead and added a check to mosh-server that causes the server to terminate after it's seen no clients for a long time.  The default timeout is one week.

Note that I'm new to both using Mosh and the source, so this might be the wrong way to do it.  Specifically:
* Should I be throwing some sort of exception?  Directly doing an exit(5) seems like a bad idea...
* Should the timeout exist in some configuration file?  I can't find a config file for Mosh so I now just have a compile-time constant.
* Is this even necessary?  I haven't tested it with the full week timeout, but I did test it with a 30 second timeout and this patch causes mosh-server to terminate when it would usually stick around.

### DIFF
--- a/src/frontend/mosh-server.cc
+++ b/src/frontend/mosh-server.cc
@@ -88,6 +88,13 @@
 
 #include "networktransport.cc"
 
+/* If no client messages are seen in this number of miliseconds, then
+ * the mosh-server process will terminate. */
+#ifndef CLEANUP_TIMEOUT
+  /* One week, in miliseconds. */
+# define CLEANUP_TIMEOUT (604800 * 1000)
+#endif
+
 typedef Network::Transport< Terminal::Complete, Network::UserStream > ServerConnection;
 
 void serve( int host_fd,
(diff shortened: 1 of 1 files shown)


jaseg:
Another way to solve this would be to present the mosh wrapper with an unique, per-client id. The mosh wrapper would then kill any existing server processes using the same id before spawning a new server.

In most cases, zombie servers are probably created when you reboot your client, and in these cases the newly spawned mosh wrapper could just kill any lingering mosh servers using the same id. This still allows multiple mosh instances per server.

A hackish way to auto-generate that unique ID could be to use a tuple of $(hostname) and $USER (thereby not allowing multiple parallell mosh sessions between two hosts without manually chaning this id).

colindean:
I was just looking for a solution to this problem. It'd be great if this made it into a release.

cgull:
See also #12

cgull:
I've merged #690 which is a more complete solution.  Let me know if this works for you.

palmer-dabbelt:
That works for me.  Thanks!

PR #589, opened 2015-03-08
Use IO::Socket::IP to prefer v6, and to not require -6 for v6-only hosts

sdstrowes:
If I'm using a host with v4 and v6, I'd like mosh to use v6 by default.

If I'm using a host with only v6, I'd like to use mosh without forcing -6.

This is a short change to /scripts/mosh to use IO::Socket::IP to achieve precisely that. -4/-6/--family semantics should be unchanged.

### DIFF
--- a/scripts/mosh
+++ b/scripts/mosh
@@ -35,7 +35,7 @@ my $MOSH_VERSION = '1.2.4a';
 use warnings;
 use strict;
 use Getopt::Long;
-use IO::Socket;
+use IO::Socket::IP;
 
 $|=1;
 
@@ -46,7 +46,7 @@ my $predict = undef;
 
 my $bind_ip = undef;
 
-my $family = 'inet';
+my $family = "unspec"; # currently unused unless overridden
 my $port_request = undef;
 
 my $ssh = 'ssh';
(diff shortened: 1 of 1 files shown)


andersk:
The current behavior is intentional and, unfortunately, currently necessary for users who connect to dual-stack hosts and then roam to a v4-only network. We cannot use IPv6 by default until we add support for v6-to-v4 roaming.

Additionally, not everyone has IO::Socket::IP installed, and Mac users seem to be having a lot of trouble trying to get it.

You can use mosh --family=unspec already, if that's the behavior you want.

sdstrowes:
Interesting. I took a look at this actually because the default AF_INET isn't happy (obviously) with v6 literals. Actually for the longest time I simply assumed mosh wasn't v6 capable, until I scratched the surface.

My usual behaviour is that I jump between dual-stack and v6-only, and with the latter v4 mosh sessions sit and wait until they have v4 again. So v6-preferred makes mosh work well for me in all environments. If/when I upgrade away from my local diffs, I guess I can alias --family=unspec!

andersk:
Yeah, we definitely want to support your use case by default eventually, but it will need to wait until we have a way to do it without breaking some existing use cases.

Issue #212, opened 2012-04-17
reconnect if server ip changed


adzeitor:
If the server changed ip address may mosh try to reconnect to the new address.

pcchou:
:+1:

mlambie:
I'm very keen for this feature too. Is it on the roadmap, or even possible?

fungs:
This is the reason why I have to restart mosh every 24 hours, so it would boost the usability for me.

NavinF:
Has this been implemented?

From the readme:

>Mosh allows the client and server to "roam" and change IP addresses, while keeping the connection alive. Unlike SSH, Mosh can be used while switching between Wi-Fi networks or from Wi-Fi to cellular data to wired Ethernet.  
  

>If the client changes IP addresses, the server will begin sending to the client on the new IP address within a few seconds.

Nippey:
Plus one. 

I have neither control over the internet connection nor over the respective hardware:
The reconnect is unpredictable somewhere during the day.

If Mosh would try to resolve my domain name, which is updated some seconds after the reconnect, I could continue to work.

osti:
I would also really appreciate this functionality. Is there any update?

fungs:
For me, there are two major show stoppers in mosh and this is one of them, if you use dynamic dns on the server side (the other being firewalls blocking mosh ports).

mlambie:
My workaround was to modify my internal DNS to always return the external IP for servers I connect to. Others could achieve the same thing with a static entry in their /etc/hosts file.

Importantly, this doesn't solve the problem of the server changing address dynamically, but it does solve the problem of a server resolving to a different (but static) IP within a set. In my case it was the IP allocated by my ISP, and a 10.0.0.0/8 address that DNS resolved for me when I was on the local network.

tangledhelix:
I'll offer another use case for this which doesn't have to do with remote networks even. I do much of my development in a local VM on my laptop, which I connect to using a dynamic DNS name (like `foo.local`, using avahi-daemon and mdns). VMware seems to like changing its IP frequently, so even though I can mosh into the VM and not have it drop my session just because I sleep the laptop, it will stop working when the VM decides to change the IP. I'm not sure of a way to make VMware stop doing this.

eskhool:
It looks like this bug will also go the way of the ssh-agent forwarding i.e. even if someone were to fork and implement it will not see the light of day from the core dev since none of them have even commented on this inspite of so many +1s

[...]

Issue #210, opened 2012-04-17
Roaming from internal to external server address


kmcallister:
Mosh doesn't allow the server to roam.  But there's one common situation where this would be extremely useful.

I have a server behind a home NAT router.  Sometimes my laptop is on my home WiFi, and I can access the server by its internal (RFC 1918) IP address.  If I'm outside my home, I instead use the NAT router's public IP, along with a different set of SSH and Mosh ports, which are forwarded to that server.  I would like to roam smoothly between these two situations.

Right now I can do it by always using the public IP, even when I'm at home.  This introduces extra network latency, but perhaps that's acceptable.  It's especially silly when the client and server share a wired Ethernet link but the Internet gateway is connected wirelessly (which is sometimes the case at my house).

In this roaming scenario, the server and client change IP address (as observed from each other) at the same time.  So I don't expect a fully automatic solution.  Perhaps we can give `mosh-client` a list of server (IP,port) pairs to try, round-robin, when packets don't seem to be getting through.

An explicit list also mitigates the DoS vulnerability of full server roaming where someone captures a legitimate packet sent by the server and retransmits it from a bogus address to make you roam there.  The attacker can only spoof an inactive server address, and perhaps the round-robin behavior limits how much service they can deny.

I'd be interested in playing with this feature, even if we don't have `mosh` wrapper support for it.

This problem is similar to the IPv6 roaming problem (#81) and the two may have similar solutions.

pcchou:
:+1:

PR #600, opened 2015-04-28
Add support for dynamic IP hosts

nogy:
Until now every time the hosts IP changed, the mosh-server would
keep running with the mosh-client unable to reconnect.
now Ctrl+^ r will re-lookup the dns entry and adjust the address
information accordingly. That way the connection can be revived.


normanr:
Instead of adding a new host arg, why not just support passing hostname or ip as the first arg?

normanr:
Actually it looks like this goes some way towards fixing #210 and #212 :-)

andersk:
@normanr: The extra argument is important to make sure we use the same IP to which we connected over SSH. If we were to re-resolve the same hostname immediately, which might get a different address family that we can’t connect over, or a different host behind a round-robin DNS load balancer.

The same concerns apply to this patch as written, but at least this will only happen if the user invokes the functionality manually. If they do, though, they might not have a good way to recover. What we really want the client to do is to try both the old and new IP addresses until it starts getting responses from the new one.

nogy:
@andersk You are right with the load balancer. The way of trying both old and new IP addresses would indeed resolve this. I was a bit afraid about the multiple resolve attempts clogging up the possibly slow connection at the times where a slow connection keeps you from getting responses from the old IP in the first place. So deciding when to start the re-lookup process automagically is what keeps me from finishing this. Manually I have a good guess on when my server changed its IP.

timofonic:
@keithw @nogy @normanr @andersk Any news about this? It's open since 2015. Why?

jkoop:
Hey just curious why this hasn't been merged, yet. Is there a UI problem? Is it compilation? Documentation? DNS round-robin?

DNS round-robin can be delt with by just mentioning (in the manual) that this feature may not work well with such domain names, since it sounds like it's totally up to the user anyway.

I'd *love* this to be merged, since my server is behind DDNS.

nogy:
Hi,

I ported this again recently. I've had a discussion with Keith about how to trigger another resolution attempt several years ago. I haven't had the time to implement the change though. (And frankly disagreed to have the resolution trigger automatically after a certain timeout). The patch gives you the option of re-resolving by a meta-keycombination. Auto-resolving would clog an already slow connection. I guess by modern standards there is no 'slow' connections anywhere but maybe in Germany *sic*. So I would add that now. On the other hand there is also a shell script wrapper that should allow you to re-connect to a mosh session iirc.

Cheers Bastian

Am 19. April 2022 23:38:24 MESZ schrieb Joe Koop ***@***.***>:

achernya:
We're going to solve this more generally with the plans in #212 and #1214. Since this has merge conflicts, closing in its current form.

PR #601, opened 2015-04-29
ipv6

ure:
mosh script requires use IO::Socket::INET6
fedora/centos/redhat rpm .spec requires perl-IO-Socket-INET6 to function


andersk:
The mosh script already uses IO::Socket::IP for IPv6 support if it is present.

ure:
Indeed my bad ;(

PR #605, opened 2015-05-18
Add git commit version info

### LINKED ISSUE #258: Git commit in mosh -v
It would be nice if `mosh{,-client,-server}` built from Git would include the output of `git describe` with the version number.

cgull:
Fixes #258.

### DIFF
--- a/scripts/Makefile.am
+++ b/scripts/Makefile.am
@@ -1,4 +1,8 @@
-EXTRA_DIST = wrap-compiler-for-flag-check
+EXTRA_DIST = wrap-compiler-for-flag-check mosh.pl
 if BUILD_CLIENT
-  dist_bin_SCRIPTS = mosh
+  bin_SCRIPTS = mosh
 endif
+CLEANFILES = $(bin_SCRIPTS)
+
+mosh:	mosh.pl ../VERSION Makefile
+	@sed -e "s/\@VERSION\@/`cat ../VERSION`/" -e "s/\@PACKAGE_STRING\@/@PACKAGE_STRING@/" $(srcdir)/mosh.pl > mosh
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,3 +1,24 @@
 ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = src scripts man conf
 EXTRA_DIST = autogen.sh ocb-license.html README.md COPYING.iOS
+nodist_noinst_HEADERS = version.h
+BUILT_SOURCES = version.h
+
+.PHONY:	VERSION
+
+VERSION:
+	@-[ -f VERSION ] || echo @PACKAGE_STRING@ > VERSION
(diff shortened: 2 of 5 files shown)


scripts/Makefile.am
```diff
@@ -2,3 +2,9 @@ EXTRA_DIST = wrap-compiler-for-flag-check
 if BUILD_CLIENT
   dist_bin_SCRIPTS = mosh
 endif
+
+mosh:	mosh.pl ../VERSION
+	@sed -e "s/\@VERSION\@/`cat ../VERSION`/" -e "s/\@PACKAGE_STRING\@/@PACKAGE_STRING@/" mosh.pl > mosh
```
glensc: why is the second replacement `@PACKAGE_STRING@` did you mean `$(PACKAGE_STRING)` ?

cgull: configure does the substitution when the makefiles are built.  PACKAGE_STRING is set in configure.ac.

andersk: `$(PACKAGE_STRING)` is still better style. Automake will automatically generate the corresponding assignment `PACKAGE_STRING = @PACKAGE_STRING@`, which allows the variable to be overridden on the `make` command line if desired.

cgull: > |$(PACKAGE_STRING)| is still better style. Automake will automatically
> generate the corresponding assignment |PACKAGE_STRING =
> @PACKAGE_STRING@|, which allows the variable to be overridden on the
> |make| command line if desired.

But if you do that, then the compile-time override is inconsistent with
the Configure-time definition of |PACKAGE_STRING| in config.h, which is
used for mosh-client and mosh-server.  That can be changed too, of
course, but is there a good reason to have a compile-time knob for this?

cgull:
I should add, I don't know that well what I'm doing here with automake, I cargo-culted this code in from someplace else. If anyone has a use case where compile time configuration for this is useful, I'll be glad to hear it.

glensc:
so, just add another commit changing it to:
```
-e "s/\@PACKAGE_STRING\@/$(PACKAGE_STRING)/" mosh.pl > mosh
```

cgull:
That much is obvious.  Less obvious is why setting PACKAGE_STRING outside of Make's dependency chain is useful.  Have you got a use case?

glensc:
dunno, just guessing, a custom vendor build:

`make PACKAGE_STRING="mosh(patched-by-seagulls)"`

cgull:
I've updated the diff but the scope remains the same-- just the git commit info.  I've gone as far into the bowels of autoconf/automake as I'm going tonight, and still don't know how to make `PACKAGE_STRING` work.  I look forward to a pull request with fully-functional `PACKAGE_STRING` functionality from one of you :)

cgull:
Pulled, plus a slight change to make mosh.pl work without edits.

PR #610, opened 2015-05-21
Small portability fixes

cgull:
This is a set of small portability fixes for OpenBSD and FreeBSD.  Derived from #523.


cgull:
Pulled.

PR #622, opened 2015-06-08
Allow non-inserting prediction (experimental).

cgull:
This adds an --overwrite option which disables prediction's visually-noisy insert/delete behavior right of the cursor, and instead relies on the server to update that text eventually.  It's not a finished work.  I spent about 5 seconds on coming up with the questionable name "overwrite" (and so I haven't documented anything yet), prediction gets turned off frequently and doesn't work as well with vertically-split windows, the update behavior on backspace is sometimes weird, etc etc.  But if people like this, all these things can be fixed.  I think it's sometimes a decent improvement, myself.

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -42,6 +42,8 @@ my $server = 'mosh-server';
 
 my $predict = undef;
 
+my $overwrite = 0;
+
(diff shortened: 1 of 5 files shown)


cgull:
Pulled with some improvements.

PR #624, opened 2015-06-09
Show the escape key properly when server unavailable.

### LINKED ISSUE #611: When the remote is unreachable, escape sequence shown doesn't obey MOSH_ESCAPE_KEY
When the client doesn't reach the server, the message shown as terminal overlay shows:

mosh: sendto: Host is down (23 s without contact.) [To quit: Ctrl-^ .]

The escape sequence shown above is hardcoded in frontend/terminaloverlay.cc:236

     const static char quit_keystroke[] = " [To quit: Ctrl-^ .]";

However the terminal sequence depends on the value of MOSH_ESCAPE_KEY environment variable and that works as expected. Only the displayed string is wrong.

The correct(ish) way to render escape key name can be found in frontend/stmclient.cc:170-188.

I don't have an opportunity to look into it. Should be a simple fix, however visibility might be a problem.

cgull:
Fixes #611.

### DIFF
--- a/src/frontend/stmclient.cc
+++ b/src/frontend/stmclient.cc
@@ -186,6 +186,7 @@ void STMClient::init( void )
     tmp = string( escape_key_name_buf );
     wstring escape_key_name = std::wstring(tmp.begin(), tmp.end());
     escape_key_help = L"Commands: Ctrl-Z suspends, \".\" quits, " + escape_pass_name + L" gives literal " + escape_key_name;
+    overlays.get_notification_engine().set_escape_key_string( tmp );
   }
   wchar_t tmp[ 128 ];
   swprintf( tmp, 128, L"Nothing received from server on UDP port %s.", port.c_str() );
--- a/src/frontend/terminaloverlay.cc
+++ b/src/frontend/terminaloverlay.cc
@@ -233,10 +233,9 @@ void NotificationEngine::apply( Framebuffer &fb ) const
     explanation = reply_message;
   }
 
-  const static char quit_keystroke[] = " [To quit: Ctrl-^ .]";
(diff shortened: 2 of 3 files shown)


cgull:
This is a small thing.  It could go into 1.2.5.  It could wait.  :)

andersk:
@cgull, can you make sure to test your commits with `./configure --enable-compile-warnings=error`?

```
  CXX      terminaloverlay.o
terminaloverlay.cc: In constructor ‘Overlay::NotificationEngine::NotificationEngine()’:
terminaloverlay.cc:166:1: error: ‘Overlay::NotificationEngine::escape_key_string’ should be initialized in the member initialization list [-Werror=effc++]
 NotificationEngine::NotificationEngine()
 ^
cc1plus: all warnings being treated as errors
Makefile:404: recipe for target 'terminaloverlay.o' failed
make[3]: *** [terminaloverlay.o] Error 1
```

keithw:
I am setting up Travis-CI now to improve this.

keithw:
Sigh, apparently there's no way to set up Travis without either moving mosh to an organization, or giving Travis access to my whole GitHub account. I'm going to revisit this later when less late.

cgull:
Once protobufs compile cleanly with clang on OS X and FreeBSD...sure!

More seriously, I'm a FreeBSD developer and, my main workstation at home is a Mac, and my main repo is there.  I use `--enable-compile-warnings=maximum`, but this error didn't appear with clang on OS X or FreeBSD.

I'll see about getting a better Linux machine set up-- the one I'm using is a near-netbook-class machine, and about 5x slower doing builds than my new Mac.

PR #625, opened 2015-06-09
1.2.4.95rc1 pre-release

cgull:
So here's build + packaging for 1.2.4.95rc1.  @keithw, would you review this?
* Ubuntu/Debian package build and install works
* Fedora build and install works
* Mac OS X package is still broken
* I'm talking to the FreeBSD ports maintainer, he seems happy to pick up the new version once we get to it

As for pushing the new packaging to Debian/Canonical and to Fedora/Redhat, who does that?  @keithw, me, Yoda?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,41 @@
+2015-06-?? Keith Winstein <mosh-devel@mit.edu>
+
+       * Version 1.2.5 released.
+
+       * New features:
+               * Bind to a specific IP address with --bind-server. (Philipp Haselwarter)
+               * MOSH_ESCAPE_KEY configures escape character.  (Timo J. Rinne)
+               * Support non-roaming IPv6. (Anders Kaseorg)
+               * Implement XTerm mouse mode. (Barosl LEE, Andrew Chin, Louis Kruger)
+               * Report Git revision along with version if available.  (John Hood)
+
+       * Platform support:
+               * Add pselect() emulation. (Jérémie Courrèges-Anglas)
+               * OpenBSD, OS X: Fix be64toh-related issues. (Jérémie Courrèges-Anglas)
+               * ARM Neon: fix gcc4.8 compiling problem(Pasi Sjöholm)
+               * NaCl: Conditionally rename main to mosh_main. (Richard Woodbury)
+               * FreeBSD: Token pasting, forkpty(), ARM fixes. (John Hood)
+               * AIX: Implement CTTY grabbing when TIOCSCTTY is missing
+                 (Anton Lundin)
+
+       * Bug fixes:
+               * Automake/autoconf workarounds.  (Anders Kaseorg)
+               * mosh-server: Allow startup without PTY.  (Keith Winstein)
+               * network.cc: Properly close old fd on Socket assignment operator.
+                 (Thanks to Igor Bukanov)
+               * mosh-server:  Allow startup with zero-window-size PTY.
+                 (Igor Bukanov)
+               * AddrInfo: Fix error message generation when node == NULL
+                 (Anders Kaseorg)
+               * Timestamp: Prevent integer overflow on Darwin PPC 32-bit
+                 (Anders Kaseorg)
+               * scripts/mosh: Fix hang when remote closes the connection
+                 (Anders Kaseorg)
+               * Fix issues with parsing of 256-color SGR sequences.
+                 (John Hood)
+               * Numerous code hygiene, Coverity, and Clang static analyzer
+                 fixes.  (Anders Kaseorg, Geoffrey Thomas, John Hood)
+
 2013-03-27 Keith Winstein <mosh-devel@mit.edu>
 
 	* Version 1.2.4 released.
--- a/configure.ac
+++ b/configure.ac
@@ -2,7 +2,7 @@
 # Process this file with autoconf to produce a configure script.
 
 AC_PREREQ([2.61])
-AC_INIT([mosh], [1.2.4a], [mosh-devel@mit.edu])
+AC_INIT([mosh], [1.2.4.95rc1], [mosh-devel@mit.edu])
 AM_INIT_AUTOMAKE([foreign -Wall -Werror])
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 AC_CONFIG_SRCDIR([src/frontend/mosh-client.cc])
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,43 @@
+mosh (1.2.4.95rc1-1) unstable; urgency=low
+
+  * Version 1.2.5 release candidate.
+
+  * New features:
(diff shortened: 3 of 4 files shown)


keithw:
Great, I will take a look at this tomorrow! Then we should email the mosh-devel and mosh-users mailing lists and ask for testing. I will also deploy the server at MIT so we can see if we hear any reports.

For packaging:

- I upload the packages to Debian unstable. I will want to upload the prerelease and make sure it builds before doing the real release.

- @achernya is the Fedora maintainer. We will want him to attempt a test build of the prerelease.

- The Ubuntu ppa:keithw/mosh-dist repo automatically tracks our "stable" ref in this repo.

- The Ubuntu ppa:keithw/mosh-dev repo tracks master (and has been tracking it).

cgull:
@achernya, sure, I'll take it out.  You're welcome to take my edits and put your name on them.

andersk:
Wait, if we’re going to include a `mosh.spec`, then it should correspond to the new version in the commit that we tag with the new version. It need not correspond with what’s in Fedora at the time the commit is made. Fedora isn’t pulling the spec file out of our VCS or anything. The primary use case for having `mosh.spec` in the repo is for non-Fedora users or people who otherwise want to build their own RPMs.

cgull:
@andersk is right, I think.  Plus, we want an up-to-date `mosh.spec` file in the prerelease so that anyone who wants/needs to test by deploying the prerelease can do so.  And there should only be a monotonic sequence of changes in that file.  So I think the answer here is for us to include `mosh.spec` changes in this and the 1.2.4 version number bumps.  Then, if Fedora requires it, @achernya can add a 1.2.4-2 version to the spec file, push that to Fedora as its initial 1.2.4 release, and commit it here.

achernya:
Fedora has no such requirement. I committed mosh.spec out of convenience for other users of mosh, but the intention was always to keep a mirror copy of Fedora. I would strongly prefer they not go out of sync; and as such, they have slightly, since I've not yet gotten around to submitting a PR with the latest changes. I'm happy to provide an updated version.

cgull:
I'm happy to take PRs from you before announcing/tagging 1.2.4.95 and 1.2.5, if that's what's needed.  Does that work for you?

andersk:
Who benefits from us keeping an out-of-date mirror of the Fedora spec? (It would be necessarily out-of-date, because the Fedora release happens strictly after our upstream release.)

Anyone looking for the Fedora spec can find it in the [usual Fedora place](http://pkgs.fedoraproject.org/cgit/mosh.git).

cgull:
Also, I am not that familiar with the RH/Fedora/RPM ecosystem.  Are there RPM distros that don't descend from Fedora?  That would indicate that our `mosh.spec` should be authoritative.

achernya:
@andersk It won't be. I will send @cgull a PR later today bringing us into sync. Plenty of other RPM based distros have used the "native" mosh spec in this repository as their starting point.

glensc:
imho in git `mosh.spec` should be renamed to `mosh.spec.in` and contain `@VERSION@` autoconf macro and thus `mosh.spec` should appear in release tarball only.

aside, pld linux (not descendant from redhat/fedora) has own .spec, and don't really care what is in mosh own repo: https://github.com/pld-linux/mosh/blob/master/mosh.spec

andersk:
Let me try to list all the ways I can think of in which our mosh.spec could work (assuming we want to keep one at all), so we can compare their advantages and disadvantages. Feel free to suggest others.

1. We release mosh-1.2.5 without a mosh.spec update; the Fedora maintainer asynchronously submits mosh.spec updates via PR to master. This means we never create our own unofficial versions of mosh.spec, but it also implies that what we tag as mosh-1.2.5 includes the spec file for 1.2.4, and so on, which seems poor.
2. We do our own unofficial version bumps in mosh.spec at upstream release time; the Fedora maintainer asynchronously submits additional updates when desired via PR to master (perhaps replacing our unofficial RPM changelog entry with the official version bump).
3. We make mosh.spec automatically generated. Users need to run the configure script to get a spec file, but otherwise, the result is essentially equivalent to (2).
4. Coordinating with the Fedora maintainer, we synchronously block Mosh releases on the corresponding Fedora releases so that the new spec file can be included.
5. We remove mosh.spec from master and put it on its own Git branch; when the new Fedora spec file is ready, the corresponding upstream release is merged into that branch, so that its contents are consistent with the spec file. Users need to checkout a separate branch to find the spec file.

I think (2) or (4) is clearly best for users. (2) is easier to maintain, but (4) may help us catch Fedora-specific problems earlier if everyone involved is willing to commit to that coordination; I don’t have a strong preference between these two. It’s not clear what would be best for RPM distributors who want to treat our mosh.spec file as authoritative (perhaps (5)?), but as far as I know, no such distributors exist.

keithw:
Re `mosh.spec`: I think we've basically been doing (4) all this time, and this is what @achernya is proposing to do here as well (send us a PR to update `mosh.spec`), so as far as I'm concerned I'm happy with the status quo. I don't think the pre-release needs to block on this though.

Should I be concerned about these make errors at the end of the compile? I assume not, but is there any way to clean this up?

      make[2]: Entering directory '/home/keithw/mosh'
      Makefile:794: recipe for target 'VERSION' failed
      make[2]: [VERSION] Error 1 (ignored)
      Makefile:802: recipe for target 'version.h' failed
      make[2]: [version.h] Error 1 (ignored)
      make[2]: Leaving directory '/home/keithw/mosh'
      make[1]: Leaving directory '/home/keithw/mosh'

keithw:
I think we have a blocking bug involving resize handling -- I am going to close this PR for now and open a new bug for it. :-(

achernya:
I normally send a sync PR for mosh.spec after the release, since the official tarball doesn't (unless this changed recently, and IMO shouldn't) contain it. The fact that the git tag includes a fedora/mosh.spec directory is a bit unfortunate, but the right way to fix that would be to move both debian/ and fedora/ to their own branches, thereby sidestepping the problem.

If you want a mosh.spec ahead of the release, I can provide it before we commit to Fedora.

keithw:
Issue at https://github.com/keithw/mosh/issues/626

keithw:
@achernya: Sounds good. I don't think it has to be perfect, but please do send a PR with your best guess of the 1.2.5 mosh.spec whenever you have it. (The more important thing is just making sure the pre-release builds and runs properly on Fedora once we have one.)

cgull:
@keithw:  I revisited the Makefile issue.  I knew it was a teeny bit wrong, but it's actually more wrong than that-- the version-file generation ignores too many errors.  I now have a fix for that, but not for the recipe being run twice (but that's harmless, since it doesn't rewrite the target if it's not changed).

cgull:
@achernya:  I looked at the mosh.spec issue a while back, and now I more fully understand how RH package specs do benefit from living separately from their source packages.  So I'll welcome a PR from you with a new spec for 1.2.5 once we get to that point.

PR #630, opened 2015-06-11
Add Travis CI configuration

andersk:
Do not pull this until Travis is enabled for this repository. Here’s the [Travis log from my own repository](https://travis-ci.org/andersk/mosh/builds/73101675).

The `git fetch --tags --unshallow` is necessary to satisfy the `VERSION` generating code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: cpp
+sudo: false
+
+addons:
+  apt:
+    packages:
+    - protobuf-compiler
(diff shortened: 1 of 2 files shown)


cgull:
`git describe --always` allows Git to report a bare hash if it can't come up with anything better.  I think I will add this to Makefile.am eventually, along with some other cleanup the version-generation stuff needs (the report's format is not consistent between Git and tarball builds, at least).

PR #641, opened 2015-06-24
Update README.md

cmcsun:
added build instructions to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,17 @@ Getting Mosh
   sure that a running process's environment variables are not readable by other
   users.  We have confirmed that this is the case on GNU/Linux, OS X, and
   FreeBSD.
+  
+Building Mosh From Source
+------------
+  On most operating systems, building is simple. Simply run the following commands as root:
+  
+  git clone https://github.com/keithw/mosh.git
+  cd mosh
+  ./autogen.sh
+  ./configure
+  make
+  make install
 
 Usage
 -----


GGMethos:
Please consider this. It took me quite some time to figure out how to build from source.

eminence:
Note that these README instructions don't cover the necessary prerequisites.  These can be found on the https://mosh.mit.edu/ home page under the "Building from source" section

I would recommend that README and the homepage contain consistent information

README.md
```diff
@@ -66,6 +66,17 @@ Getting Mosh
   sure that a running process's environment variables are not readable by other
   users.  We have confirmed that this is the case on GNU/Linux, OS X, and
   FreeBSD.
+  
+Building Mosh From Source
+------------
+  On most operating systems, building is simple. Simply run the following commands as root:
```
andersk: It not recommended to build software as root. Only the `make install` step needs root access (and then only if you are installing to a root-owned directory such as the default `/usr/local`).

The-Compiler: Even when doing that it's a very bad idea to run `make install` as root. You're installing software system-wide by hand - only the package manager can do that, so you can actually manage (e.g. uninstall) said software.

It's usually much better to let the package manager install the compiled files - e.g. `checkinstall` on Debian, writing a small `PKGBUILD` on Archlinux, etc.

cgull:
I'd suggest that the build directions ought to be in only one place.  I think the thing to do might be to place build instructions only in README.md, then link to it on Github from mosh.mit.edu.  Moving the directions off the homepage also makes the choice less attractive to less-experienced admins.

eminence:
I now am thinking that the instructions should be in both locations (web and README.me), and they should be consistent/identical.  I admit the possibility of them diverging over time due to neglect doesn't make me happy, but in practice these instructions change so rarely that ought not to be a real problem.

achernya:
I believe this has been made obsolete by #1252

Issue #650, opened 2015-07-23
Address system() call portability


paralin:
The `locale` command is not available on all machines.

It's totally unnecessary to run `system("locale")` on line 316 of mosh_server.cc. I vote to remove it.

keithw:
Arguments for keeping it:

(1) Mosh only tries to run it if the user's configuration is already broken and we need to print a diagnostic message. The diagnostic message needs to be verbose and explanatory because this is a confusing area, Mosh is more pedantic than most programs, and nobody wants to believe that their configuration is actually broken.

(2) Having "locale" is a POSIX/Single Unix Spec requirement (since 2001). Every POSIX system has it.

(3) A missing locale is a nonfatal error. (And mosh is already quitting already when we get to that point!)

(4) If a system really doesn't have it, the distributor could remove it on that architecture.

Arguments for getting rid of it:

(1) Which platforms are we talking about that don't have it?

jmcp:
On 24/07/15 06:50 AM, Christian Stewart wrote:
> The |locale| command is not available on all machines.
>
> It's totally unnecessary to run |system("locale")| on line 316 of
> mosh_server.cc. I vote to remove it.

ugh .. execing system() is a generally bad idea. Use of
execve() is much better.


James C. McPherson
--
Oracle
Systems / Solaris / Core
https://www.jmcpdotcom.com/blog

glensc:
the code bloat around execve() proper handling (fork() etc) is not worth of it. if you use system() with constant parameters (not coming from any user input), it's not  that big security issue.

paralin:
The system that doesn't have it is an ODROID XU3 running a Buildroot system which uses Busybox as the core. 

I don't think executing a system() call just to output some debug information is justified. The user already knows they have a locale problem, does running `locale` really output valuable enough information to not just tell the user to go do it themselves?

I'm writing a buildroot cross-compilation package for Mosh at the moment which is why I'm dealing with these edge cases. I think in something SSH-like and security-oriented like Mosh, portability should probably be emphasized. Calling the `locale` command seems really against that.

keithw:
Yes, in practice, we have had a lot of user confusion around mismatched locales (and Mosh's strict enforcement of a working configuration before starting up), which made us make this error message more and more verbose way back in Mosh 1.2 (released April 2012).

We often have to ask users to paste the full error message so we can diagnose the problem. The "locale" section has been helpful to see what is going on, e.g. which vars are being set explicitly and which are inferred!

[...]

PR #651, opened 2015-07-23
Remove system("locale") calls.

paralin:
The locale command is not available on many systems. As this variable
is unused and appears to have been written with the intent of
displaying the locale settings to the user, it's not really necessary.
As this breaks Mosh on a lot of systems, it's best to remove the calls.


keithw:
Please see comment at https://github.com/keithw/mosh/issues/650#issuecomment-124238062

Hard for me to believe this "breaks Mosh on a lot of systems."

Mosh only attempts running the command as part of a fatal error message, a missing command would be a nonfatal error, and the "locale" command is required by POSIX (since 2001).

Where are you seeing a problem?

paralin:
See #650

PR #653, opened 2015-07-24
Better, more flexible handling for IPv4/IPv6.

cgull:
This PR is a try at improving the usability of mosh's current "non-roaming" IPv6 support.  It adds automatic protocol selection for hostnames or addresses that are unambiguously IPv4 or IPv6, and a I-don't-care option for people who use Mosh in a fully dual-stack environment.  It also attempts to connect to all available addresses for the server, which mosh was missing support for.

Since the code has to enumerate all of a hostname's addresses to do all this, it might also be a good base for future work to support true dual-stack roaming, at least at session setup.

This isn't quite done yet.  I've had some more late ideas, like a `--family=prefer-ipv4` and `--family=prefer-ipv6` to allow either protocol but strongly prefer only one.  More ideas are welcome.

### DIFF
--- a/man/mosh.1
+++ b/man/mosh.1
@@ -64,6 +64,11 @@ pass between client and server. By default, \fBmosh\fP uses the ports
 between 60000 and 61000, but allows the user to request a particular
 UDP port instead.
 
+Currently, \fBmosh\fP has limited support for IPv6, dual-stack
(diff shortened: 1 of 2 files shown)


cgull:
I've committed an improved version of this.

PR #668, opened 2015-08-21
Make --help or -h work again

eminence:
At the moment, I don't think there are any codepaths that will print the mosh-server usage information.  This commit fixes that by explicitly handling --help and -h

### DIFF
--- a/src/frontend/mosh-server.cc
+++ b/src/frontend/mosh-server.cc
@@ -101,9 +101,9 @@ static int run_server( const char *desired_ip, const char *desired_port,
 
 using namespace std;
 
-static void print_usage( const char *argv0 )
+static void print_usage( FILE *stream, const char *argv0 )
 {
-  fprintf( stderr, "Usage: %s new [-s] [-v] [-i LOCALADDR] [-p PORT[:PORT2]] [-c COLORS] [-l NAME=VALUE] [-- COMMAND...]\n", argv0 );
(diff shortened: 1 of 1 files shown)


andersk:
Conventionally, programs invoked with `--help` or `--version` are expected to write to stdout and exit with status 0 (see, for example, the [GNU coding standards](http://www.gnu.org/prep/standards/standards.html#g_t_002d_002dhelp)). Yeah, the `mosh` script violates this right now, but it shouldn’t.

eminence:
Good point.  I've updated the PR to now `exit( 0 )`

andersk:
`print_usage` still goes to stderr. It should go to stdout on `--help` and stderr in other cases.

eminence:
Ack, sorry.   Let's try again!

andersk:
I wonder if we want to use a real `getopt_long` parser here (perhaps importing the Gnulib module for platforms without `getopt_long`), rather than this ad-hoc strategy that mishandles theoretical cases like `mosh-server -i --help`?

eminence:
Sure, that seems reasonable.  But I would propose that be handled in a different PR.

eminence:
Does anything else need to be done for this fix to be merged?

cgull:
I don't think so. But I seem to be the only person actively working on Mosh, and I've been working on a test framework + tests as prelude for bringing in my performance stuff, and I don't have a ton of time to spend on Mosh.

I'll get to it eventually.  If you can clone me, that'll help...

Issue #81, opened 2012-03-25
Missing IPv6-support


skibbi:
Mosh does not seem to support IPv6. When specifying a hostname with both IPv4 and IPv6, an IPv4-connection is used. Specifying an IPv6-address results in:

/usr/bin/mosh: Could not resolve hostname 2a01:238:42a0:a200:e709:8d1c:581d:6e5a
ssh_exchange_identification: Connection closed by remote host
/usr/bin/mosh: Did not find remote IP address (is SSH ProxyCommand disabled?).

dtaht:
will mosh multipath be merged?

cgull:
Not for mosh-1.2.5, which is just a rollup of bugfixes and cleanup PRs that were in the queue-- no more new features.  I haven't even looked at https://github.com/boutier/mosh yet, though I'm definitely interested.

nomadium:
@cgull is there an ETA for mosh-1.2.5?

cgull:
I should have the two remaining issues wrapped up this weekend-- OS X build and Debian bash completion.  Release will come soon after that, depending on how much bake time it gets and if we find anything else.

nomadium:
Good to know, thanks for taking care of the release.

kurtkraut:
With the mosh -6 parameter I'm able to connect to my server using IPv6 but without mentioning this parameter, mosh (version 1.2.5 tested) will always prefer to connect using IPv4 instead of IPv6 if a AAAA DNS entry is available for the FQDN.

I strongly suggest as browsers, OpenSSH and other client/daemon softwares behave, AAAA is prefered over A addresses.

Mikaela:
> I strongly suggest as browsers, OpenSSH and other client/daemon softwares behave, AAAA is prefered over A addresses.

They follow [RFC 6555](https://tools.ietf.org/html/rfc6555) aka [Happy Eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs) which everything should support nowadays.

graingert:
@Mikaela but mosh doesn't need to support Happy Eyeballs, it can make multiple connections simultaneously

dtaht:
I really would like to see mosh multipath merged. It's even happier than happy eyeballs.

pavlix:
I hope you don't mind that I created an issue just for the roaming, see #855 . This is an old issue for lack of IPv6 support. I have tried using mosh 1.2.6 on IPv6 with success and therefore I believe this issue should just be closed.

[...]

PR #670, opened 2015-08-26
Validate bind-server option for both IPv4/6

robbat2:
bind-server option needs to support both IPv4 and IPv6 addresses, as it
previously only tried to detect IPv4 addresses. Now uses regex to
validate the addresses fully, including support for scoped IPv6
link-local in validating the IPv6 address.

Fixes: 669
Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -162,12 +162,8 @@ if ( not defined $bind_ip or $bind_ip =~ m{^ssh$}i ) {
   push @bind_arguments, '-s';
 } elsif ( $bind_ip =~ m{^any$}i ) {
   # do nothing
-} elsif ( $bind_ip =~ m{^[0-9\.]+$} ) {
-  push @bind_arguments, ('-i', "$bind_ip");
 } else {
-  print STDERR qq{$0: Unknown server binding option: $bind_ip\n};
-
-  die $usage;
+  push @bind_arguments, ('-i', "$bind_ip");
 }
 
 if ( defined $fake_proxy ) {


andersk:
Yikes. Can we just delete this validation code instead? The user is not the enemy here, and the address is already validated inside `mosh-server` by `getaddrinfo` with `AI_NUMERICHOST` in `ai_flags`.

robbat2:
Redone

normanr:
Works for me.  Although I can't think of a situation (with the current head) where binding to different IPv6 address (different from the SSH_CONNECTION) would be useful.  Multipath (as mentioned in #81) would definitely benefit  (but it's not merged at head yet).  Dynamic re-resolve (re: #600) I would only expected to apply to IPv4 addresses (unless there are crazy ISP's that dynamic v6 subnet assignments).

andersk:
(For the record: I don’t know if Comcast counts as a crazy ISP, but it does account for the world’s [highest IPv6 traffic volume](http://www.worldipv6launch.org/measurements/), and it does dynamic IPv6 subnet assignments. We are also worried about split-horizon DNS environments with IPv6 on one side and IPv4 on the other.)

normanr:
Yea, I guess bind to :: would be useful if you have split dns (especially split v4/v6) and reresolve patch applied.

cgull:
`--bind-server` turns out to be important for IPv6 link-local addresses.  Which is how I discovered that I hadn't actually pulled this fix.  Pulled for real now.

PR #676, opened 2015-09-24
e2e test framework

cgull:
Work in progress.  Very sausage-y.  Don't look too closely-- I'm just letting Travis tenderize this.


cgull:
The real PR is #679.

PR #679, opened 2015-10-03
finished e2e test framework

cgull:
This is a new end-to-end test suite for Mosh sessions, along with a few changes to Mosh to support it.  This also adds a Travis OS X build.  `make distcheck` is successful on OS X 10.9, FreeBSD 10.2, Ubuntu 14.04, and Ubuntu 15.10.  See src/tests/README.md.

Supersedes #676, which I was using for development.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,10 @@
+os:
+  - linux
+  - osx
+
+cache:
+  - apt
+
 language: cpp
 sudo: required
 services: docker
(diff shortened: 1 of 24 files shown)


keithw:
@andersk, do you have a view on this? It looks pretty good to me.

cgull:
It looks half baked to me :)  Overgrown shell scripts, lack of real test framework/architecture, etc, etc.  But it's better than what we had before...

But I've added one more test that's important for my performance code, and I'm getting to a point where I'd like to merge the PR.  Comments?

.travis.yml
```diff
@@ -1,3 +1,10 @@
+os:
+  - linux
+  - osx
+
+cache:
+  - apt
```
The-Compiler: FWIW, this won't do anything for free Travis accounts with `sudo: required`.

cgull: I already had some question about that.  Our Travis Linux build specification is I think fairly mixed up about what it's trying to do.

I just remembered that the standard build supports fetching from custom repos and Launchpad PPAs, we can probably get a tmux-=>1.8 that way.

PR #682, opened 2015-10-06
Switch to Nettle crypto lib-- trial balloon

cgull:
This was motivated by pain on OS X 10.11's dropping OpenSSL headers but not dylib.
This would probably want to be a configurable option, but it demonstrates that it's fairly easy for Mosh to switch crypto libraries.  There's also the idea of using Apple's Common Crypto on OS X.


cgull:
Travis whines because I didn't switch the dependencies in its config.  Ignore that and look at the code.  :)

cgull:
Closed in favor of #686.

PR #690, opened 2015-11-02
Server timeouts

cgull:
I've had this sitting around a while.  Discussion on #12 is making me post it now...

This adds environment variables to optionally control session timeout/termination behavior on mosh-server.

* `MOSH_SERVER_NETWORK_TMOUT` controls how long mosh-server will wait after last hearing traffic before exiting.
* `MOSH_SERVER_SIGNAL_TMOUT` controls how long mosh-server will wait after last hearing traffic before allowing SIGUSR1 to be effective.

`MOSH_SERVER_NETWORK_TMOUT` is pretty straightforward and obvious.  `MOSH_SERVER_SIGNAL_TMOUT` is less so.  The objective there is to allow simple scripts to kill off network-idle sessions without risk of killing sessions that do have connectivity.  I'm not sure how much this is needed and how easy it is for people to grasp and use.  I could drop that part entirely, or tie the SIGUSR1 functionality to MOSH_SERVER_NETWORK_TMOUT instead.

### DIFF
--- a/man/mosh-server.1
+++ b/man/mosh-server.1
@@ -92,33 +92,32 @@ effect.
 
 .TP
 .B MOSH_SERVER_NETWORK_TMOUT
-When set to an integer number, this variable specifies how long the
-server will wait, in seconds, to receive an update from the client
-before exiting.  Since \fPmosh\fP is very useful for mobile clients
-with intermittent operation and connectivity, we suggest setting this
-variable to a high value, such as 604800 (one week) or 2592000 (30
-days).  When not set, \fBmosh-server\fP will wait indefinitely for a
-client to reappear.  This variable is somewhat similar to the
-\fBTMOUT\fP variable found in many Bourne shells.  However, it is not
-a login-session inactivity timeout; it only applies to network
-connectivity.
+If this variable is set to a positive integer number, it specifies how
+long (in seconds) \fBmosh-server\fP will wait to receive an update from the
+client before exiting.  Since \fPmosh\fP is very useful for mobile
+clients with intermittent operation and connectivity, we suggest
+setting this variable to a high value, such as 604800 (one week) or
+2592000 (30 days).  Otherwise, \fBmosh-server\fP will wait
+indefinitely for a client to reappear.  This variable is somewhat
+similar to the \fBTMOUT\fP variable found in many Bourne shells.
+However, it is not a login-session inactivity timeout; it only applies
+to network connectivity.
 
 .TP
-.B MOSH_SERVER_NETWORK_TMOUT
-When set to an integer number, this variable specifies how long the
-server will wait, in seconds, to receive an update from the client
-before allowing \fBSIGUSR1\fP to terminate a session.  When unset,
-\fBSIGUSR1\fP will always terminate sessions.  With this feature,
-users and administrators may implement scripts to clean up
-disconnected Mosh sessions without fear of accidentally killing those
-that have active connectivity.  With this variable set, a user
-connected via Mosh can
-issue
+.B MOSH_SERVER_SIGNAL_TMOUT
+If this variable is set to a positive integer number, it specifies how
+long (in seconds) \fBmosh-server\fP will ignore SIGUSR1 while waiting
+to receive an update from the client.  Otherwise, \fBSIGUSR1\fP will
+always terminate \fBmosh-server\fP.  Users and administrators may
+implement scripts to clean up disconnected Mosh sessions.  With this
+variable set, a user or administrator can issue
+
 .nf
 $ pkill -SIGUSR1 mosh-server
 .fi
-to kill disconnected sessions without killing their current login
-session.
+
+to kill disconnected sessions without killing connected login
+sessions.
 
 .SH EXAMPLE
 
--- a/man/mosh-server.1
+++ b/man/mosh-server.1
@@ -82,6 +82,27 @@ Locale-related environment variable to try as part of a fallback
 environment, if the startup environment does not specify a character
 set of UTF-8.
 
+.SH ENVIRONMENT VARIABLES
+These variables allow server-side configuration of Mosh's behavior.
+They may be set by administrators in system login/rc files,
+/etc/login.conf, or similar mechanisms, or users in their shell's
+login/rc files.  \fBmosh-server\fP passes these variables to the login
+session and shell that it starts, but changing them there will have no
+effect.
+
+.TP
+.B MOSH_SERVER_NETWORK_TMOUT
+When set to an integer number, this variable specifies how long the
+server will wait, in seconds, to receive an update from the client
+before exiting.  Since \fPmosh\fP is very useful for mobile clients
+with intermittent operation and connectivity, we suggest setting this
+variable to a high value, such as 604800 (one week) or 2592000 (30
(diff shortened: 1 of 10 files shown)


cgull:
The reason I chose environment variables:
* Administrators can easily set it as a default
* Users can set it to what they want, whether or not admins have done anything
* Users can override admins' defaults
* Somewhat analogous to `TMOUT`

But I'm willing to hear arguments for other channels.

man/mosh-server.1
```diff
@@ -82,6 +82,44 @@ Locale-related environment variable to try as part of a fallback
...
+with intermittent operation and connectivity, we suggest setting this
+variable to a high value, such as 604800 (one week) or 2592000 (30
+days).  When not set, \fBmosh-server\fP will wait indefinitely for a
+client to reappear.  This variable is somewhat similar to the
+\fBTMOUT\fP variable found in many Bourne shells.  However, it is not
+a login-session inactivity timeout; it only applies to network
+connectivity.
+
+.TP
+.B MOSH_SERVER_NETWORK_TMOUT
```
infinity0: I guess this should be `MOSH_SERVER_SIGNAL_TMOUT`?

cgull: Yes, of course.

man/mosh-server.1
```diff
@@ -82,6 +82,44 @@ Locale-related environment variable to try as part of a fallback
...
+When set to an integer number, this variable specifies how long the
+server will wait, in seconds, to receive an update from the client
+before allowing \fBSIGUSR1\fP to terminate a session.  When unset,
+\fBSIGUSR1\fP will always terminate sessions.  With this feature,
+users and administrators may implement scripts to clean up
+disconnected Mosh sessions without fear of accidentally killing those
+that have active connectivity.  With this variable set, a user
```
infinity0: On a quick skim I mis-interpreted this as "only users connected via Mosh can [..]". Perhaps instead of "with this variable set, a user [..] can issue", say "Another use case is to allow a user [..] to issue".

src/frontend/mosh-server.cc
```diff
@@ -545,14 +572,26 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
...
-      int timeout = min( network.wait_time(), terminal.wait_time( now ) );
+      timeout = min( timeout, network.wait_time() );
+      timeout = min( timeout, terminal.wait_time( now ) );
       if ( (!network.get_remote_state_num())
 	   || network.shutdown_in_progress() ) {
         timeout = min( timeout, 5000 );
       }
+      /* The server goes completely asleep if it has no remote peer. */
+      if ( network_timeout_ms ) {
+	int64_t wait = network_timeout_ms - ( now - network.get_latest_remote_state().timestamp );
```
infinity0: `wait` is already a function (in `sys/wait.h`). It doesn't cause problems currently, but it might be better to call this `wait_ms`.

src/frontend/mosh-server.cc
```diff
@@ -336,6 +339,26 @@ int main( int argc, char *argv[] )
...
 		       const string &command_path, char *command_argv[],
 		       const int colors, bool verbose, bool with_motd ) {
+  /* get network idle timeout */
+  long network_timeout = 0;
+  char *timeout_envar = getenv( "MOSH_SERVER_NETWORK_TMOUT" );
+  if ( timeout_envar ) {
+    /* We don't care about strtol errors here, they all produce sensible return values. */
+    network_timeout = strtol( timeout_envar, NULL, 10 );
+    if ( network_timeout < 0 ) {
+      network_timeout = 0;
```
infinity0: If there is an error, this value gets set to 0, but in line 718 below 0 is treated as "stay alive indefinitely". This might bite users - they intended something else, but aren't told about the error. Granted, it looks like this is exactly what bash does with `TMOUT` but it's not a good principle to follow. One solution is to have 0 actually mean "exit immediately". If you don't want to do that, you should at least make the docs more precise:

- When set to an integer number -> When set to a positive integer number
- When not set -> when not set, or set to an invalid or non-positive value

cgull: You have a good point about the silent error.  But having a bad value mean "exit immediately" results in another problem (at least for bash and `TMOUT`):  if you pass a bad value from your `.profile`, then your shell will exit immediately next time you login, and you will have an interesting time fixing that.  It's less of an issue here, since the timeout only operates on lost network connectivity, but we might then surprise people by operating differently than `TMOUT`.  I'll think about this one.

ayust:
:+1:

I wouldn't mind seeing a command-line option on the default client that makes it easier to set the value (i.e. the command-line option just causes the client to invoke mosh-server with the corresponding environment variable set), but that's just a nice-to-have level feature, unlike the presence of the capability at all.

(Setting this from the client side can make sense when some endpoints are more likely to shut down uncleanly than others - e.g. a machine which is often rebooted while off the network, versus a machine which has longer uptime but intermittent connectivity.)

src/frontend/mosh-server.cc
infinity0: You could have mosh-server output a warning to mosh-client whenever the timeout is infinite. That would also cover the case where the user doesn't know that the default is infinite. (I'd still argue for a finite default, e.g. 30 days, but not sure if you guys are so keen on that idea.)

cgull:
@ayust, There's nothing explicit, but something like
`mosh --server='env FOO=BAR /usr/local/bin/mosh-server'  host`
will do what you want.

ayust:
@cgull yeah, just requires a more intimate knowledge of mosh's working. (Fine for me, maybe not for another user.)

eminence:
+1 for me (on the idea, I've not reviewed the code yet).   I have argued in the past (and still am) that `MOSH_SERVER_NETWORK_TMOUT` is a useful mechanism for dealing with dead clients (clients that will never come back online)

cgull:
Rebased and with man page and error message improvements.  I chose to leave Mosh's behavior as it is now, partly for backward compatibility with existing usage, and partly because setting a fixed default timeout is a policy, and this feature is (and I think should be) a mechanism.

src/frontend/mosh-server.cc
```diff
@@ -549,14 +582,31 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
...
   while ( 1 ) {
     try {
+      static const uint64_t timeout_if_no_client = 60000;
+      int timeout = INT_MAX;
       uint64_t now = Network::timestamp();
 
-      const int timeout_if_no_client = 60000;
-      int timeout = min( network.wait_time(), terminal.wait_time( now ) );
+      timeout = min( timeout, network.wait_time() );
+      timeout = min( timeout, terminal.wait_time( now ) );
```
The-Compiler: I don't think it's intended to set `timeout` twice here, right?

cgull: Note the min() on both lines.  The intent is to find the lowest timeout value from several possible wait periods.  I believe it's correct.

The-Compiler: Ah, indeed - sorry, I need more coffee! :wink:

infinity0:
LGTM, haven't run it myself yet though.

infinity0:
Er, I take that back. `strtol` is stupid. Added some more comments.

infinity0:
Looks good!

Issue #499, opened 2014-02-07
Support for reverse port forwarding


wincent:
This would be a nice to have; basically the equivalent of SSH's `-R` option:

```
 -R [bind_address:]port:host:hostport
         Specifies that the given port on the remote (server) host is to be forwarded
         to the given host and port on the local side.  This works by allocating a
         socket to listen to port on the remote side, and whenever a connection is
         made to this port, the connection is forwarded over the secure channel, and
         a connection is made to host port hostport from the local machine.

         Port forwardings can also be specified in the configuration file.  Privi-
         leged ports can be forwarded only when logging in as root on the remote
         machine.  IPv6 addresses can be specified by enclosing the address in square
         brackets.

         By default, the listening socket on the server will be bound to the loopback
         interface only.  This may be overridden by specifying a bind_address.  An
         empty bind_address, or the address `*', indicates that the remote socket
         should listen on all interfaces.  Specifying a remote bind_address will only
         succeed if the server's GatewayPorts option is enabled (see sshd_config(5)).

         If the port argument is `0', the listen port will be dynamically allocated
         on the server and reported to the client at run time.  When used together
         with -O forward the allocated port will be printed to the standard output.
```

My specific use case here is wanting to use https://github.com/wincent/clipper, which relies on reverse port-forwarding to expose the local clipboard to a process running on a remote host (ie. to do things like copy from tmux or Vim on a remote machine and have it land in your local system clipboard).

I've seen tickets here for agent forwarding and "normal" (not reverse) port-forwarding, so thought I'd open up a ticket to track this one separately.

xrchz:
:+1:

alxyang:
👍

pbjc:
👍

d4l3k:
duplicate of #337

wincent:
@d4l3k: I created this one as a separate issue specifically for *reverse* port forwarding, because I think people are talking about *forward* port forwarding in that other issue.

d4l3k:
That other issue mentions both the -R and -L flag.

wincent:
Ya, I know, but the features are conceptually distinct. I'd hate to see forward port forwarding get implemented (because it is more obvious) and have reverse port forwarding overlooked.

ayr-ton:
It could be just a wrapper around SSH as well. Just for enabling replacing SSH by Mosh in VSCode SSH environments for example (It needs SSH forwarding for working as expected)

PR #423, opened 2013-05-14
Out-of-band data + SSH Agent forwarding

rinne:
Pluggable out of band communication mechanism over Mosh transport
layer and agent forwarding support in top of out of band mechanism.

I contrubute this code to Mosh project under the same license that
Mosh itself is distributed. All licensing options and clauses
included.

Signed-off-by: Timo J. Rinne <tri@iki.fi>

Issue #416, opened 2013-04-12
Blend sshuttle like TCP tunneling


leleobhz:
https://github.com/apenwarr/sshuttle

The idea is do data-over-UDP more or less like sshutle does:

"But you can't safely just forward TCP packets over a TCP session (like ssh), because TCP's performance depends fundamentally on packet loss; it must experience packet loss in order to know when to slow down! At the same time, the outer TCP session (ssh, in this case) is a reliable transport, which means that what you forward through the tunnel never experiences packet loss. The ssh session itself experiences packet loss, of course, but TCP fixes it up and ssh (and thus you) never know the difference. But neither does your inner TCP session, and extremely screwy performance ensues.

sshuttle assembles the TCP stream locally, multiplexes it statefully over an ssh session, and disassembles it back into packets at the other end. So it never ends up doing TCP-over-TCP. It's just data-over-TCP, which is safe."

If this kind of code can be implemented in mosh, TCP packages can be assembled in client and data-over-UDP can be sent. Fixes issue with TCP-over-TCP and make things faster using UDP.

Issue #337, opened 2012-10-21
ssh port forwarding doesn't work


staticfloat:
I believe this is because of the fact that `mosh-server` takes over from `sshd` and `sshd` quits, therefore the tunnel disappears just after it is made.  It would be nice if mosh could keep `sshd` alive if `-L` or `-R` are present in the `--ssh` argument

omry:
why wouldn't mosh be able to reconnect the tunnel as well?

brainstorm:
Yeah, I would like to drop those lines in my `~/.ssh/config`:

```
Host somehost
	HostName example.com
	LocalForward 5558 example.com:5558
	User admin
```

And be able to have that local tunneling done with mosh too. Is it that different from a plain SSH connection that drops and recovers?

brainstorm:
I opened a $50 bounty for this issue:

https://www.bountysource.com/issues/4471419-ssh-port-forwarding-doesn-t-work

Should anyone solve this one or related ones like:

https://github.com/mobile-shell/mosh/issues/120

dwwoelfel:
The bounty for this feature is up to $200, now.

mharradon:
I threw in a bit, up to $235.

Could anybody familiar with the Mosh codebase comment on the expected difficulty of this addition? Perhaps point out the relevant files? That might make it easier for somebody to take a crack at it.

wittrup:
I currently use some ugly "_auto_"ssh script that run ssh like this:

```bash
reverse=$(( ((RANDOM<<15)|RANDOM) % 63001 + 2000 ))
/usr/bin/ssh -o ExitOnForwardFailure=yes$identity_file -N -R $reverse:localhost:22 -p $port $host; } 2>&1
```

Having mosh do that for my network of raspberry pi would be awesome.  Bounty increased.

claui:
The bounty has now reached $300.

MisterTea:
FYI: port forwarding works in Eternal Terminal with the -t option.  https://mistertea.github.io/EternalTCP/

brainstorm:
@MisterTea sure, but my HPC cluster has only mosh server(s) installed so far... I reckon that with ET they would have to install yet another service? If ET works well against Mosh server, that would solve the issue for me, if not, this bounty is still relevant to me ;)

MisterTea:
@brainstorm Yep, ET is it's own server that you have to install.  Where I work we have 100s of engineers using ET, but it's up to your use case to decide whether it works well for you.

[...]

PR #696, opened 2015-11-28
SSH agent forwarding patches re-merged as one patch.

rinne:
Hi

SSH agent forwarding patches re-merged as one patch.

A couple of superficial conflicts resolved. Tested and working.

Merged as requested in #583 but since #583 was made by someone else, I decided just to make another PR.

Signed-off-by: Timo J. Rinne <tri@iki.fi>


rinne:
Hi

I wonder what causes the autobuild failure. It seems to be that one protocol buffers generated header fails to include. Compiles fine for me in OpenSUSE Leap and OS-X. Must be something trivial. Maybe someone else can give it a quick look. (probably Makefile.am, src/agent/Makefile.am, or src/network/Makefile.am is to blame).

cgull:
It fails for me on OS X with `make distcheck`, but not `make check`.  This suggests an issue with your Makefile.am files, probably stemming from the way `make distcheck` uses separate source and build directories.  Try `make distcheck V=1`.

rinne:
Fixed!

rinne:
Perhaps should not have squashed my trivial Makefile.am fix directly to previous commit, but that's exactly what I did :/. Anyways, now it works and it the stuff is still in one commit. The original arrangement was to have OOB protocol first and agent stuff on top of that in another commit, but I suppose no-one is about to use OOB to anything else besides agent, so I can't imagine OOB being merged but agent forwarding omitted.

src/agent/agent.cc
```diff
@@ -0,0 +1,490 @@
...
+    int i;
+    for ( i = 0; i < 10; i++ ) {
+      dir += voc.substr( prng.uint32() % voc.length(), 1 );
+    }
+    if ( mkdir( dir.c_str(), 0700 ) != 0 ) {
+      return;
+    }
+    string path(dir + "/");
+    for ( i = 0; i < 12; i++ ) {
+      path += voc.substr( prng.uint32() % voc.length(), 1 );
```
daniel-pittman: This is really not even remotely secure.  It would be better to use http://pubs.opengroup.org/onlinepubs/9699919799/functions/mkdtemp.html instead.

src/frontend/mosh-server.cc
```diff
@@ -502,6 +517,14 @@ static int run_server( const char *desired_ip, const char *desired_port,
       exit( 1 );
     }
 
+    /* set SSH_AUTH_SOCK */
+    if ( agent.active() ) {
+      if ( setenv( "SSH_AUTH_SOCK", agent.listener_path().c_str(), true ) < 0 ) {
+	perror( "setenv" );
```
daniel-pittman: given multiple setenv calls, it would be nice to upgrade to `setenv(SSH_AUTH_SOCK)` here, I think, and the corresponding change in, eg, the following perror.

src/network/outofband.cc
```diff
@@ -0,0 +1,323 @@
...
+
+OutOfBand::OutOfBand()
+  : comms(),
+    datagram_instruction_out(),
+    reliable_instruction_out_sent(),
+    reliable_instruction_out_unsent(),
+    seq_num_out( 0 ),
+    ack_num_out( 0 )
+{
+  seq_num_out = 0;
```
daniel-pittman: that duplicates the initializer list...

src/agent/agent.cc
rinne: It would be nice to have a bit more elaborate reference to some insecure bit. Claiming "not even remotely safe" needs a bit more than just a statement of "fact".

What it tries to do, is to fail in case the directory can't be safely created, which makes a single agent forwarding fail in _extremely_ unlikely case. While using mkdtemp(3) would probably be ok, the problem is that it doesn't get mode as argument and changing the mode afterwards would easily create a race condition and changing mosh server umask doesn't feel very appealing. Historically unix domain socket endpoints have not been particularly safe on all platforms (they should be safe in 2010's systems though) and that's why they are in security apps wrapped into directories.

That much said, if someone likes to modify the patch to use mkdtemp, I don't object.

src/network/outofband.cc
rinne: Patch to remove duplicate is welcome.

src/frontend/mosh-server.cc
rinne: I didn't really understand, but please patch it a way you like.

rinne:
Merged current upstream master branch and resolved a couple of trivial conflicts.

pjjw:
just wanna shout my <3 for this PR.

src/agent/agent.cc
tamird: Sorry, what is the race condition created by mkdtemp?

tribut: Just for completeness: According to the [the man page](http://linux.die.net/man/3/mkdtemp), `mkdtemp` always creates directory with permissions `0700`.

rinne: Quickly glancing through current glibc version of mkdtemp(3) and it seems fine. As said earlier, I don't mind if someone wants to convert (and converts) the patch to use mkdtemp. It's just that I won't be using my time doing that.

rinne: Actually remembering the past and the ssh server implementation, the most tricky thing in creating user's agent proxy socket was due to two things, neither of which is present in mosh.

1) The socket was created while the server was still running with root privileges.

2) All sockets for a single user were generated into a same directory (like /tmp/ssh-keithw, later in openssh those paths are "anonymized" so another user can't as easily create a directory to tmp blocking other user).

Point 1 is obvious in mosh, because mosh server is executed wth user credentials.

Point 2 is a double bladed sword. It surely makes the code more complex and in some cases may even allow other local users to block agent forwarding from the others. On the other hand, each new connection doesn't create another directory to tmp that needs to be cleaned up later.

In mosh patch, the simplest approach is taken and new directory is created for each proxy socket and  path name is generated with enough random so that collisions are extremely improbable (and while they may occur in principle, it would only mean that agent forwarding for that particular connection would not work).

tamird: I do not understand this comment as a justification for rolling your own `mkdtemp`. It seems to me that cleaning up directories created by `mkdtemp` and friends is not necessary, since they will eventually be cleaned up automatically.

rinne: I think you misunderstood. It was not justification of anything. Just a brief history. And should you be up to it, you are welcome to submit a patch that makes the code use mkdtemp(3). It's done now like it is, just because of a faint recollection that some implementations of mk*t*mp were broken sometime around the dawn of time and instead of looking into it further I just wrote it like I did. Should there be a bug, please point it out. Should you want to get rid of it in favour of mkdtemp(3), write and submit a patch, and stop whining.

rinne: And absolutely, it's not ok, to create a new directory to /tmp with every login and leave it there "because it's going to be eventually cleaned up". That's absolutely idiotic, no matter how the directory gets created.

toresbe:
This makes me so happy. Thank you very much!

src/frontend/mosh-server.cc
werkshy: I think this comment is suggesting we could define a constant `SSH_AUTH_SOCK` to use in the `setenv()` calls instead of a hardcoded string.

```
#define SSH_AUTH_SOCK "SSH_AUTH_SOCK"
```

rinne: Well, since SSH_AUTH_SOCK is well known environment variable and can't be changed because it's what ssh uses, there isn't much point in defining it. getenv("SSH_AUTH_SOCK") i.m.o. is easier to write and understand than getenv(SSH_AUTH_SOCK). All the latter provides with, is a possibility to break the code in a subtle way by changing the value of the defined constant. It's almost as useful as:

```
#define HOME_DIRECTORY_ENV_NAME "HOME"
...
getenv(HOME_DIRACTORY_ENV_NAME)
...
```

magicmark:
Hmm, tried this but getting:

~ ❯ mosh markl@xxxxxx --forward-agent
new: invalid option -- 'A'
Usage: mosh-server new [-s] [-v] [-i LOCALADDR] [-p PORT[:PORT2]] [-c COLORS] [-l NAME=VALUE] [-- COMMAND...]

I'm pretty sure I installed it right, but it doesn't seem to work :(
Any thoughts?

rinne:
@magicmark Yes. Also server must be compiled from agent-forwarding-merge-20151128 branch using
--enable-agent-forwarding configure option and installed so that it is the server that gets executed when you log in.

jedahan:
Would really love to see this merged, is there anything blocking it that we could help?

yannk:
I'm curious about what others are doing to work around the lack of agent forwarding in Mosh. Personally, aside from waiting for this to merge, I'm using a parallel ssh connection when I need my agent to be running on the remote; predictable SSH_AUTH_SOCK location helps with sharing it with Mosh when I need to resume my work flow in my Mosh session.

glensc:
@yannk i run [screen](https://www.gnu.org/software/screen/) on machine i mosh to, and so the agent runs there, i also use [keychain](http://www.funtoo.org/Keychain) to keep `SSH_AUTH_SOCK` inherited properly. i also have custom scripts that handle replacing $SSH_AUTH_VALUE with a symlink to working ssh auth sock when SSH_AUTH_SOCK gets lost under keychain (some cron wipes time to time /tmp sockets). i recently found [byobu](http://byobu.co/) which seems to point SSH_AUTH_SOCK to it's run dir ($BYOBY_RUN_DIR/sockets) having it managed properly

jrgifford:
Is there a reason that this isn't merged yet? Anything we can do?

pedro-araujo:
I would like to use this patch as well. Is it not merged yet due to security implications?

arronmabrey:
@rinne looks like there are merge conflicts again. :-(

rinne:
@arronmabrey Once again merged and resolved (trivial'ish) conflicts. I won't do those forever though ...

arronmabrey:
@rinne yeah, it's a shame, they're hanging you out to dry so long with this PR. ☹️

cgull:
I'd like to consider #120/#583/this for Mosh 1.3, after we release 1.2.6.  I've started a discussion on the mosh-devel mailing list; if you're interested please comment there.

mathias-lang-sociomantic:
Any update on this ? AFAICS Mosh v1.2.6 was released in August. Like many people, this is one of the most annoying issue I have with mosh.

Your [mailing-list discussion](http://mailman.mit.edu/pipermail/mosh-devel/2016-May/001297.html) did not seem to have encounter any opposition.

rinne:
@mathias-lang-sociomantic I don't know. Once again I did a manual merge so now it again should be in sync with upstream master branch.

rinne:
@Mic92 I think I won't bother any more. If there are conflicts and I feel like it, I once and awhile merge and resolve them, but that's it. They should just merge it to mainline and be done with it.

andersk:
@Mic92 It's easy enough to squash the changes to a single patch if/when this gets accepted, and meanwhile you can consult the Changes tab of the PR to see the single patch. Presentation issues are not the blocker here.

src/frontend/mosh-server.cc
hschmidt: I agree, extracting a constant or macro gives the illusion of variability where there is none.

hschmidt:
Wow, a PR that's been open for a year with nothing but cosmetic issues, debatable ones at that. What is going on?

andersk:
@hschmidt It sounds like you misread my comment. Cosmetic issues are _not_ the blocker here.

hschmidt:
@andersk Sounds good. Keep up the good work.

rinne:
@hschmidt "... PR that's been open for a year ..."

Actually I wrote bulk of the agent forwarding for Mosh in April-May 2013 so its 4 year birthday is quickly approaching. It just lives in a newer PR instead of the original one which was #423.

It took free nights of a week or two to put that together and occasional 30 to 60 minutes after that and it has well paid off for me, because I'm using it all the time. I haven't been holding my breath while waiting to see this in mainline for a last couple pf years, but I'm certain it would be useful.

hschmidt:
@rinne, yeah, I've been following this agent forwarding thing since 2012. I envy your persistence, seriously! I know we can't expect the the same *user support* as we do from commercial projects, but when it comes to handling *user contributions* open source maintainers may want to apply a slightly higher standard as far as commitment is concerned.

In case anyone is looking to call me a hypocrite, I have multiple GitHub accounts: I'm also @hannes-ucsc and @eyealike.

cgull:
I also have some concerns about the implementation, and @keithw has voiced some concerns about how the ssh-agent messaging attaches to the existing network stack.  So someone has to spend some time with the code to work through these issues, and I've been saying I want to, but it hasn't happened yet.

rinne:
The expressed concerns have been so generic that it is not very easy to address them. However I'm quite sure that things like "I think it attaches the network on the different level I'd like it to attach" is something that is not going to change for two very clear reasons:
1. Guys involved certainly don't use and depend on agent functionality. If they did, the feature would already be in, because for people who are using it, it is hugely important and useful.
2. Moving the implementation to another layer of protocol stack is more or less complete rewrite of the functionality, because most of the patch is passing the agent data back and forth through the protocol stack. The reason why this rewrite is never going to take place is threefold:
  a. I'm not going to bother.
  b. No-one outside mosh core developers is going to bother remembering that 4 years has passed with this one.
 c. No-one inside mosh core development is going to bother because of point 1 above.

Just reserve the protobuf message numbers that are used by agent patch in the mainline so that no-one will accidentally reuse them for something else. Other than that, I really don't care one way or another. Nor will I care about random concerns, unless they at some point become specific.

infernix:
This is quite a sad state of affairs. I think there are many users like me who tried mosh and then abandoned it solely because of the lack of ssh-agent forwarding; I've been subscribed to this and previous issues/PRs in anticipation of a merge for a long time now. There are 100+ participants over at https://github.com/mobile-shell/mosh/issues/120

Could the mosh team please at least consider spending a bit of time on formulating their objections so that there's a chance of moving this into mainline?

ErwanAliasr1:
+1 for the sshfs over mosh

thapakazi:
**lets get it merged please team..**.
lets get it merged

rlpowell:
Is anyone actively maintaining a public fork that has this integrated?  I mean, that seems like the obvious solution here.

rinne:
@rlpowell The branch from which this pull request has been made, is with regular inregularity synced with upstream master branch. That's what I use myself. Server and client must both be complied from this branch and configured with --enable-agent-forwarding flag.

https://github.com/rinne/mosh/tree/agent-forwarding-merge-20151128

kuozumijp:
v1.3 Released... 
Why aren't merged yet?

jrgifford:
No conflicts, all checks have passed... :\

apenney:
Every time I come to check the status of this I get sad.  This is the single thing that stops me recommending mosh to my entire team, and spreading it through organizations I work with.  Almost every one of them uses a bastion host.

tnguyen14:
Could the maintainer of mosh seriously consider merging this, based on the feedback of users? I've been subscribed to this thread and #120 for a few years now, and I have not heard many complaints as to why this shouldn't be in. It's been an overwhelming plea to get this merged and released so we can use mosh more broadly.

bukzor:
I've started a bounty for this bug, because money talks, and put in $50 to get it rolling, I hope.

https://www.bountysource.com/issues/1587475-ssh-agent-forwarding

@cgull: Have a cookie!  🍪

jedahan:
I wonder if this is the record for longest open merge-able PR (especially considering this is the second attempt).

bukzor:
IMO the best remaining solution is a rename of @rinne's branch. Perhaps `mosh-A`, referencing the -A agent-forwarding option of ssh. Then I can make a homebrew package and get the chrome-mosh extension to use it, etc. It's difficult / impossible to do those things when the two binaries have the same name and one is a "feature branch" of the other. Homebrew and the chrome-mosh maintainer won't buy it, in that setting :(

geohuz:
Hi all, I just found this thread and tried compiled source on server and client, changed my .ssh/config to:
```
Host *
  RemoteForward 9997 127.0.0.1:9999
```

and mosh to server with agent-forward option, but I didn't see any process running on server listen to the 9997 port

how can I fix the issue?

andersk:
@geohuz At this time, Mosh does not support port forwarding of any kind, with or without this PR—see #337, #416, #499. This PR is specifically about [https://en.wikipedia.org/wiki/Ssh-agent](SSH agent) forwarding, not port forwarding.

nikhilweee:
Can someone take the initiative to describe the current state of affairs to a newbie like me? Will this allow me to use sshfs over mosh? If yes, why hasn't this been merged yet?

andersk:
@nikhilweee No. This patch is about [SSH agent](https://en.wikipedia.org/wiki/Ssh-agent) forwarding and is not relevant to any other kind of forwarding (port, filesystem, X11, or otherwise). We have other open feature requests for those kinds of ideas. An sshfs analogue has been requested as #280, but as far as I know, nobody has tried to make it into anything more than an idea.

If you don’t have a specific requirement for SSH agent forwarding (you would know it by that specific name if you do), this PR is not what you’re looking for.

arekm:
Is anyone publishing standalone patches for released mosh versions with changes from this branch/pull req?

Mic92:
I have rebased the pull request on top of current master:

https://github.com/Mic92/mosh/commits/agent-forwarding

kuozumijp:
Why isn't it merged yet?

MichaelErmer:
After reading most of the discussions (https://github.com/mobile-shell/mosh/issues/120) and reading about Guardian Agent (https://github.com/StanfordSNR/guardian-agent), I think you should re-consider merging this PR, as we live in different times now. 

For me and many others, my local "ssh-agent" is 1password or another password/secret manager, which asks every time my ssh key will be used if I want to allow that. It is an improved Guardian Agent. 
Bildschirm­foto 2023-02-05 um 11 54 54

Besides that, if people want to spread their insecure ssh-agents, it's up to them, there may be good reasons for them not to care, but I don't think the software should enforce something on them.

The lack of agent forwarding in mosh forces me to now, after years of enjoying mosh, use ssh since I started using a password/secret manager instead of local ssh keys and copying them around computers and internet-exposed contexts (servers).

lroehrs:
Also need this patch, without is mosh for us useless. :/

lroehrs:
Also need this patch, without is mosh for us useless. :/

nyawox:
now that official bitwarden-desktop client has implemented ssh agent support this has become more relevant than ever. (although it's still behind feature flag and not active yet in their saas instance, i'm already daily driving in my private instance)
I understand the concerns behind forwarding ssh agent at that time, and probably even i would have made the same decision. But the situation has changed. All my ssh keys are stored in a encrypted vaultwarden database, which I plan to harden further by rotating the database credentials using hashicorp vault or something. And my ssh-agent replaces sudo password prompt. This is arguably more secure than reusing the same memorable user password across all deployed hosts.

PR #708, opened 2016-01-04
Debian package updates

andersk:
Removes an extra `-- "$@"` from debian/mosh.maintscript that was causing errors; silences a warning about unused `${perl:Depends}`; and allows building in parallel.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -5,8 +5,8 @@ Maintainer: Keith Winstein <keithw@mit.edu>
 Build-Depends: debhelper (>= 7.0.50), autotools-dev, protobuf-compiler, libprotobuf-dev, dh-autoreconf, pkg-config, libutempter-dev, zlib1g-dev, libncurses5-dev, libssl-dev, bash-completion
 Standards-Version: 3.9.6.1
 Homepage: http://mosh.mit.edu
-Vcs-Git: git://github.com/keithw/mosh.git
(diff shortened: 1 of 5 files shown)


debian/control
```diff
@@ -5,8 +5,8 @@ Maintainer: Keith Winstein <keithw@mit.edu>
 Build-Depends: debhelper (>= 7.0.50), autotools-dev, protobuf-compiler, libprotobuf-dev, dh-autoreconf, pkg-config, libutempter-dev, zlib1g-dev, libncurses5-dev, libssl-dev, bash-completion
 Standards-Version: 3.9.6.1
 Homepage: http://mosh.mit.edu
-Vcs-Git: git://github.com/keithw/mosh.git
-Vcs-Browser: https://github.com/keithw/mosh
+Vcs-Git: https://github.com/mobile-shell/mosh.git
```
ghostbar: why change it from git to https?

andersk:
(This was already merged as 9cf54e388e836b8cf950885bb353ee6d5a8595b8, 8816493ec9fcaadceb10e313493c3eec08c08212, f928435274e4e1325daa90fc051650115f0256b8, c9ec54a004fababacca4a83e010c13fb0edb2bdf.)

debian/control
andersk: git:// is not secure; https:// is.

PR #709, opened 2016-01-07
updated bash completion

sbdchd:
I modified the bash completion file so that flags are partially completed.

### DIFF
--- a/conf/bash-completion/completions/mosh
+++ b/conf/bash-completion/completions/mosh
@@ -1,9 +1,18 @@
+#!/bin/bash
+
 _mosh () {
-    local cur prev
+    local simple_flags="-a -b -4 -6 -p"
+    local flags="--client= --server= --predict= --family= --port= 
+    --bind-server= --ssh= --no-init --help --version"
 
-    _init_completion || return
-
-    _known_hosts_real -a "$cur"
+    local cur=$(_get_cword)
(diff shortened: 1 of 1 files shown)


cgull:
We want this, I think, but I'm no bash-completion expert.
* What OSes and versions have you tested this on?
* What versions of bash-completion and Bash will this work with?

sbdchd:
I have tested this with bash_completion 2.1 & bash version 4.3 on Ubuntu and OS X.

I am fairly certain that this requires at least bash_completion 2, since it uses `_init_completion`.

cgull:
The thing that's keeping me from taking this is that OS X has bash 3.2 as its default shell, which these changes will break on.  But at least with our packages and Homebrew, bash-completion is not configured by default, and so I don't *think* that's actually a problem.  I don't know anything about MacPorts or enough about bash-completion usage on OS X to decide-- anyone have any wisdom in this area?

sbdchd:
The existing implementation of bash-completion for mosh requires at least bash-completion 2, which itself requires Bash 4.

So the current setup doesn't work on OS X without installing bash-completion 2 and upgrading the default Bash installation to Bash 4.

cgull:
I just merged a combination of your PR and @y-higuchi's bash-completion compatibility code.  Would you please verify that this works as you expect?

sbdchd:
Working great!
Thanks.

cgull:
Pulled.

PR #716, opened 2016-01-17
Use server's $SSH_CONNECTION if ProxyCommand isn't run.

cgull:
This is an alternate way to resolve #417 without affecting
backward compatibility.


cgull:
Pulled, and already replaced by better code.

Issue #727, opened 2016-03-10
SIGSEGV in mosh-server on debian sid


gibbon-joel:
Hi,

I guess I need a little help to provide useful information on this, but I have a problem with the mosh-server process not starting up properly on debian sid. I noticed the problem first yesterday (9.3.) morning when trying to login into my server.

Starting the process manually on the affected system yields:

$ mosh-server

MOSH CONNECT 60001 <OTP>

mosh-server (mosh 1.2.5) [build mosh 1.2.5]
Copyright 2012 Keith Winstein <mosh-devel@mit.edu>
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

[mosh-server detached, pid = 3321]
$

which looks ok. But there is no mosh-server process running and a message in the kernel log:
[ 4097.845386] mosh-server[3321]: segfault at 0 ip           (null) sp 00007fff59cd0fa8 error 14 in mosh-server[55715b883000+59000]

last lines of strace -f mosh-server:
....
ioctl(5, TIOCSPTLCK, [0])               = 0
ioctl(5, TCGETS, {B38400 opost isig icanon echo ...}) = 0
ioctl(5, TIOCGPTN, [1])                 = 0
stat("/dev/pts/1", {st_mode=S_IFCHR|0620, st_rdev=makedev(136, 1), ...}) = 0
open("/dev/pts/1", O_RDWR|O_NOCTTY)     = 6
ioctl(6, TIOCSWINSZ, {ws_row=58, ws_col=272, ws_xpixel=0, ws_ypixel=0}) = 0
clone(strace: Process 3333 attached
child_stack=0, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fc0096e2a10) = 3333
[pid  3333] set_robust_list(0x7fc0096e2a20, 24) = 0
[pid  3332] close(6)                    = 0
[pid  3332] rt_sigaction(SIGCHLD, {SIG_DFL, [], SA_RESTORER|SA_RESTART, 0x7fc008d8bd30}, {SIG_DFL, [], 0}, 8) = 0
[pid  3332] --- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0} ---
[pid  3333] close(5)                    = 0
[...]

PR #732, opened 2016-03-19
configure.ac:  Do s/-pthreads -lpthreads/-pthreads/ for protobuf

### LINKED ISSUE #730: Unable to connect since recently
I've installed mosh on two Arch Linux machines, both with fully up-to-date software. As of the last few weeks (?), I've been unable to use mosh to remotely access either one from the other, even though plain SSH works just fine. Even if both the laptop and desktop are sitting next to each other on the same desk, and connected via an ethernet switch, any time I run mosh I get the blue bar with "No response received from server" message, before mosh-client exits.

It used to work—the relevant ports in iptables are open on both machines (UDP 60000:61000 and TCP 22)—and disabling iptables has no effect. I'm not sure what else to check to see if I've misconfigured something somewhere, or whether it's a mosh/SSH/iptables/Linux bug.

If anyone can suggest things for me to test, I'd be happy to try it out, as this is confusing me and quite annoying—the main reason I was using mosh was because it would reestablish the connection after the laptop had been suspended, which I do every day to take it to class.

cgull:
protobuf uses an obsolete automake pthreads detection macro,
which results in "pkgconfig --libs protobuf" returning
"-lprotobuf -pthread -lpthread" on Linux.  Remove
the unnecessary and dangerous -lpthread in that case.

Fixes #730, mosh-server crash in libutempter on Debian Sid.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -446,6 +446,13 @@ AC_CHECK_DECL([IUTF8],
 # Checks for protobuf
 PKG_CHECK_MODULES([protobuf], [protobuf])
 
+# On Debian Sid 2016-03-16ish at least, "pkgconfig --libs protobuf"
+# gives us the bogus "-lprotobuf -pthread -lpthread"; remove
+# "-lpthread" which misconfigures compile, causing a segfault in
+# mosh-server.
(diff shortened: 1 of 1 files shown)


cgull:
Oops, this is supposed to be for #727.

Issue #117, opened 2012-04-10
Put the server name (not IP) somewhere in the mosh-client command line


nelhage:
If I connect to a server via `mosh HOSTNAME`, the mosh wrapper script spawns a `mosh-client` process, passing the IP address of the target server on the command line. If I have multiple mosh sessions, and later want to kill or debug or `reptyr` that process, I have to find it  by IP in the process-table, rather than hostname. It would be cool if the long-running `mosh-client` process had the hostname somewhere in `argv`, so that it was easy to recognize in `ps` or `top`.

PR #742, opened 2016-04-02
Display original command line argv via no-op flag.

netj:
Improves 679b819216e1946ebe3a6f920c0fbf61d61ab47e (for #117) to make it simpler to ignore mosh-client from OS X Terminal.app's "Ask before closing" process list. (See: http://superuser.com/a/254647)

Moreover, the command displayed from ps is a completely functional one, which was not the case in the previous approach, that mangles `argv[0]`.

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -422,7 +422,7 @@ if ( $pid == 0 ) { # child
   $ENV{ 'MOSH_KEY' } = $key;
   $ENV{ 'MOSH_PREDICTION_DISPLAY' } = $predict;
   $ENV{ 'MOSH_NO_TERM_INIT' } = '1' if !$term_init;
-  exec {$client} ("$client @cmdline |", $ip, $port);
+  exec {$client} ("$client", "-# @cmdline |", $ip, $port);
 }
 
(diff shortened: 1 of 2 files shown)


cgull:
Having a separate argv[0] is fine, and I agree it's more correct than what we were doing.  This will also help with `pkill` and other things that look at.  (We might as well construct a full argv, too, instead of having argv[1] be concatenated arguments.)

But what does the `-#` argument gain?  How does having a "completely functional" command help you?  (And it doesn't work for me, and I'm not often invoking `mosh-client` by copy pasta to begin with.)

netj:
Thanks for taking a look.  The emphasis of the PR was on keeping `argv[0]` intact to play it nice with OS X's Terminal.  Your first paragraph wasn't clear to me whether the PR was a good idea or not.

I learned from commit history that the reason behind rewriting `argv[0]` was to display the original hostname in addition to its IP from `ps` listing, etc.  The only other way to expose such original args of `mosh` on `mosh-client` is to keep them on a flag whose argument is ignored, hence the new `-#`.  The "completely functional" part of my comment was just a minor remark, and I agree nobody wants to run `mosh-client` commands.

PR #744, opened 2016-04-05
configure.ac: link test for -fstack-protector-all

baruchsiach:
There are (broken) compilers out there that accept -fstack-protector-*
parameters, but do not provide the needed ssp library. As a result, build does
not fail, but link does. Use the local check_link_flag autoconf function to
correctly test for SSP availability.


cgull:
This change looks fine and I will merge it, but I'd like to know what environment (compiler, OS, architecture) you encountered this problem on, so it's a concrete bugfix rather than an abstract improvement.

baruchsiach:
The Imagination Technologies provided Codescape IMG cross toolchain for the MIPS architecture version 2015.10 has this issue.

PR #746, opened 2016-04-27
Clarify that -p does not affect SSH port

ashmew2:
Please review these changes. In case you think something's out of place, let me know and I will fix it right up. Thank you.


cgull:
I pulled this a while back.

PR #752, opened 2016-05-10
GCC 6 fixes.

### LINKED ISSUE #719: Mosh protobufs create code who indentation causes a warning on GCC 6
Flagged by Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=811579

The new warning seems to be in `-Wall`. Our package builds with `-Werror`, so this breaks the package build on GCC 6.

Our options seem to be: (1) give up on -Wall for building the protobufs, (2) give up on -Werror for building the protobufs, (3) get protobuf-compiler to produce better code, and get that new protobuf-compiler released and included in the Debian stretch release.

cgull:
Should resolve #719 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=811579

This is my current thinking for the `-Wmisleading-indentation -Werror` issue revealed there.  Unfortunately, gcc 4.2.1 does not support `-Wno-error=misleading-indentation` (because it doesn't have `-Wmisleading-indentation`), and other non-GCC compilers may also get upset about it (clang is OK though).  The C++ code in this directory is all generated, not hand-authored, and `-Wno-error` seems an acceptable compromise to me.  If anyone else has better ideas, or wants to contribute `autoconf`-fu to this problem, speak up!

### DIFF
--- a/src/protobufs/Makefile.am
+++ b/src/protobufs/Makefile.am
@@ -1,7 +1,7 @@
 source = userinput.proto hostinput.proto transportinstruction.proto
 
 AM_CPPFLAGS = $(protobuf_CFLAGS)
-AM_CXXFLAGS = $(WARNING_CXXFLAGS) $(HARDEN_CFLAGS) $(MISC_CXXFLAGS)
+AM_CXXFLAGS = $(WARNING_CXXFLAGS) $(HARDEN_CFLAGS) $(MISC_CXXFLAGS) -Wno-error
 
 SUFFIXES = .proto .pb.cc
 
--- a/src/tests/encrypt-decrypt.cc
+++ b/src/tests/encrypt-decrypt.cc
@@ -54,7 +54,7 @@ const size_t NUM_SESSIONS         = 64;
 
 bool verbose = false;
 
-#define NONCE_FMT "%016"PRIx64
+#define NONCE_FMT "%016" PRIx64
 
 static std::string random_payload( void ) {
   const size_t len = prng.uint32() % MESSAGE_SIZE_MAX;


keithw:
lgtm. Did the protobuf people end up agreeing to fix this upstream? (I realize this doesn't help us for the near term...)

cgull:
Doesn't help for the lifetime of Ubuntu 16.04 LTS :)

cgull:
I don't think I ever reported this on protobuf, did you or anybody else?

They haven't moved yet on the `-lpthread` issue I opened, though.

keithw:
I guess I confused this with the `-lpthread` thing.

keithw:
There was a ton of discussion here: https://gcc.gnu.org/ml/gcc-bugs/2016-01/msg03043.html

I wonder if the warning is still generated on current GCC 6. It looks like maybe they had a change of heart but I haven't tested.

cgull:
It's still failing on

`gcc version 6.0.0 20160414 (experimental) [trunk revision 234994] (Ubuntu 6.0.1-1ubuntu11)`

which I *think* is a gcc trunk version but I'm not 100% sure.

If so, it would seem to postdate the discussion at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69415 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812005, a very similar compile fail to ours (which has a report of success with a newer compiler).

cgull:
I finally found my `sid` VM hiding under the couch cushion.  It still breaks the compile with
`gcc version 6.1.1 20160510 (Debian 6.1.1-2)` so I've pulled this change.

PR #757, opened 2016-05-23
Release 1.2.5.95rc1

cgull:
I think these are the necessary changes for release-- just changelogs and version numbers.  @achernya, @andersk, @keithw, would you review?

This is building fine on my usual platforms (Ubuntu, Raspbian, OS X, FreeBSD) and it seems to be packaging/installing fine on Debian Sid, Fedora 22, OS X Homebrew, and OS X native package build.  It looks like a much less painful release than 1.2.5, but I'm probably missing several elephants standing around.


xtaran:
A few minor nitpicks on the Debian packaging/build:

* The `Homepage` header in `debian/control` is set to http://mosh.mit.edu/, but redirects to https://mosh.mit.edu/, so probably should be set it directly to https://mosh.mit.edu/.
* The same counts for the link in `debian/copyright` to http://www.gnu.org/licenses/: Probably also should be changed to https://www.gnu.org/licenses/ to avoid the unencrypted redirect.
* Standards-Version could be set to `3.9.8`. Only difference to `3.9.7` affects `debian/*menu` files and mosh doesn't have one, so the mosh packaging already complies with 3.9.8 and needs no other changes.
* `build-package.sh` and `macosx/build.sh` get executable bits during the build, hence showing up as changes in the git repository. Maybe do a `chmod a+x build-package.sh macosx/build.sh` and commit it to git to silence this.
* There's also one "pedantic"-level lintian warning, but I assume that's on purpose for still being able to build on Ubuntu 10.04 LTS Lucid, at least according to https://github.com/mobile-shell/mosh/commit/c3fbad29f9efbbab1e5ebe31f4f8369c63794e67 from 2012:
```
P: mosh source: package-uses-old-debhelper-compat-version 7
N: 
N:    The debhelper compatibility version used by this package is marked as
N:    not recommended by the debhelper developer. You may consider using a
N:    recommended compatibility version.
N:    
N:    The compatibility version can be set in (preferred) debian/compat or by
N:    setting and exporting DH_COMPAT in debian/rules. If it is not set in
N:    either place, debhelper defaults to the deprecated compatibility version
N:    1.
N:    
N:    Refer to the debhelper(7) manual page for details.
N:    
N:    Severity: pedantic, Certainty: certain
```

debian/changelog
```diff
@@ -1,3 +1,54 @@
...
+    * Fix issue with RECVTOS error message on client on FreeBSD.
+      (John Hood)
+
+  * Bug fixes:
+    * Remove an assertion causing aborts on Unicode fallback found by
+      fuzzing with afl.  (Keith Winstein)
+    * Fix a server hang with XON/XOFF on BSD systems.  (John Hood)
+    * Fix a typeahead-prediction bug that caused display corruption on
+      urxvt.  (John Hood)
+
```
andersk: This entry is missing a date line, and dpkg complains about this.

cgull:
First, thanks to both of you for the comments!

@andersk: I'm a little embarrassed, thanks for the catch.  I'm unfortunately not surprised that got by me, I spent 2 unpleasant hours arguing with VirtualBox and disk-space issues on a Fedora VM before I got around to sid.  I checked it much better this time.

@xtaran:  Not all the license texts in `debian/copyright` are ours, they are legal documents, and my first instinct is to leave them unaltered (at least those that are not ours).  But this is an area I'm not that familiar with.  Is there any consensus within Debian or elsewhere about how these may/may not be altered?

And how are `build-package.sh` and `macosx/build.sh` not already executable for you?  They are in my work dir, and logs don't indicate any historical changes to their permissions in the repo.  Are you using a native/local Linux filesystem for your work?

On `debhelper`:  10.04 LTS is out of support now, and we all have the 3 following LTS releases to choose from.  On the other hand, the warning is thoroughly pedantic since newer releases will have a newer debhelper anyway.  I'm slightly inclined to move to 9.20120115 (12.04 and wheezy), and let people suffering with older releases backport, but if anyone has a strong opinion, say it.

(As always, I am gobsmacked by the complexity of Debian packaging.)

achernya:
@cgull I've travelling this week and won't have a chance to look in detail until Saturday. Proposed mosh.spec changes LGTM; I'll submit a PR for any changes I need to make to the specfile after I upload the release.

cgull:
Once people like what they see here, I plan to push this to master as 1.2.5.95rc1, get packages built and announce it, and give people a week or so to test it.  So we don't need to get Fedora sorted Right Away.

xtaran:
Hi,

On Mon, May 23, 2016 at 08:03:54PM -0700, cgull wrote:
> @andersk: I'm a little embarrassed, thanks for the catch.  I'm
> unfortunately not surprised that got by me, I spent 2 unpleasant
> hours arguing with VirtualBox and disk-space issues on a Fedora VM
> before I got around to sid.  I checked it much better this time.

Oh, I thought, it was on purpose to show it's not yet released yet or
so. Will argue about that next time, too. ;-)

> @xtaran:  Not all the license texts in `debian/copyright` are ours,
> they are legal documents, and my first instinct is to leave them
> unaltered (at least those that are not ours).

Good point. But then again in case of GPL version, debian/copyright is
only a pointer -- as is that URL.

(Actually there's also a discussion ongoing at Debian if the "Format:"
field is still correct if changed to https:// and if the
debian/copyright format standard should be altered to make https://
part of the standard.)

> But this is an area I'm not that familiar with. Is there any
> consensus within Debian or elsewhere about how these may/may not be
> altered?

Well, the Debian URL ChecKer (DUCK) started complaining about such
URLs which redirect to https:// in general. But as mentioned above
it's not yet completely undisputed.

> And how are `build-package.sh` and `macosx/build.sh` not already
> executable for you?

git showed them as "modifed".

> Are you using a native/local Linux filesystem for your work?

Yes, ext4. So I assume, you don't. MacOS X maybe?

Will check back later and maybe send you a pull request against the
release-1.2.6 branch for that.

> On `debhelper`:  10.04 LTS is out of support now, and we all have
> the 3 following LTS releases to choose from.

Argh. I mixed up 10.04 with 12.04. Of course, 10.04 is no more
supported.

> On the other hand, the warning is thoroughly pedantic since newer
> releases will have a newer debhelper anyway. I'm slightly inclined
> to move to 9.20120115 (12.04 and wheezy),

If you don't use a specific feature, you don't need to use a specific
debhelper version. "debhelper (>= 9)" suffices totally then.

> (As always, I am gobsmacked by the complexity of Debian packaging.)

There are lot's of things which can be done, indeed. And you can solve
a lot of non-trivial issues with it. :-)

		Kind regards, Axel
-- 
/~\  Plain Text Ribbon Campaign                   | Axel Beckert
\ /  Say No to HTML in E-Mail and News            | abe@deuxchevaux.org  (Mail)
 X   See http://www.nonhtmlmail.org/campaign.html | abe@noone.org (Mail+Jabber)
/ \  I love long mails: http://email.is-not-s.ms/ | http://abe.noone.org/ (Web)

cgull:
On 5/24/16 3:13 AM, Axel Beckert wrote:
>> @xtaran: Not all the license texts in `debian/copyright` are ours,
>> they are legal documents, and my first instinct is to leave them
>> unaltered (at least those that are not ours).
> 
> Good point. But then again in case of GPL version, debian/copyright is
> only a pointer -- as is that URL.
> 
> (Actually there's also a discussion ongoing at Debian if the "Format:"
> field is still correct if changed to https:// and if the
> debian/copyright format standard should be altered to make https://
> part of the standard.)
> 
>> But this is an area I'm not that familiar with. Is there any
>> consensus within Debian or elsewhere about how these may/may not be
>> altered?
> 
> Well, the Debian URL ChecKer (DUCK) started complaining about such
> URLs which redirect to https:// in general. But as mentioned above
> it's not yet completely undisputed.

OK, then I think I will not change them for now.

>> And how are `build-package.sh` and `macosx/build.sh` not already
>> executable for you?
> 
> git showed them as "modifed".
> 
>> Are you using a native/local Linux filesystem for your work?
> 
> Yes, ext4. So I assume, you don't. MacOS X maybe?
> 
> Will check back later and maybe send you a pull request against the
> release-1.2.6 branch for that.

Yeah, a PR might help illustrate this.  I'm still unclear on how this is
happening for you.  What are you doing to clone the repo?  What are you
doing to build?  What error messages are you getting?  When does git
show them as modified?

>> On `debhelper`: 10.04 LTS is out of support now, and we all have
>> the 3 following LTS releases to choose from.
> 
> Argh. I mixed up 10.04 with 12.04. Of course, 10.04 is no more
> supported.
> 
>> On the other hand, the warning is thoroughly pedantic since newer
>> releases will have a newer debhelper anyway. I'm slightly inclined
>> to move to 9.20120115 (12.04 and wheezy),
> 
> If you don't use a specific feature, you don't need to use a specific
> debhelper version. "debhelper (>= 9)" suffices totally then.

I have no insight on debhelper development, and the 9.<date> versioning
doesn't describe the significance of changes.  I thought choosing the
version wheezy uses seemed a reasonable target.

regards,

  --jh

xtaran:
Hi,

On Tue, May 24, 2016 at 11:24:46AM -0700, cgull wrote:
> >> And how are `build-package.sh` and `macosx/build.sh` not already
> >> executable for you?
> > 
> > git showed them as "modifed".

When preparing that PR, I noticed that it's actually vice versa and it
shows them as changed from 755 to 644:

```
→ git diff build-package.sh  macosx/build.sh
diff --git a/build-package.sh b/build-package.sh
old mode 100755
new mode 100644
diff --git a/macosx/build.sh b/macosx/build.sh
old mode 100755
new mode 100644
```

That probably explains why it looks right for you. But it doesn't make
any sense to me. So I wonder even more what made that change.

> I'm still unclear on how this is happening for you. What are you
> doing to clone the repo?

I did `git clone -b release-1.2.6 git@github.com:cgull/mosh.git` (copy pasted from my shell history).

```
→ git version
git version 2.8.1
```
  
> What are you doing to build?

IIRC this:

```
→ autoreconf -f -i
→ debuild -uc -us -j16 -b
```

(`autoreconf` is needed to get the `configure` script and other expected files. `debuild` is a wrapper around `dpkg-buildpackage` which logs all output
and runs `lintian` afterwards.)

> What error messages are you getting?

None that I'm aware of. Just the `git diff` output shown above.

> When does git show them as modified?

After the package build.

But anyways, forget about this. It would have made sense to fix it if it would have been the other way round as I initially thought.

> > If you don't use a specific feature, you don't need to use a specific
> > debhelper version. "debhelper (>= 9)" suffices totally then.
> 
> I have no insight on debhelper development, and the 9.<date> versioning
> doesn't describe the significance of changes.

Yes, IMHO date based versioning is an epidemic plague these days. [Semantic Versioning](http://semver.org/) makes much more sense to me.

> I thought choosing the version wheezy uses seemed a reasonable
> target.

IMHO that restricts the build-dependency unnecessarily.

cgull:
I still can't reproduce this on Sid.  And `macosx/` isn't even in the distribution; autoconf/automake don't know about it at all.  So a Debian build shouldn't be touching `macosx/build.sh` at all.  My guess is you have some bit of local strangeness going on.

cgull:
@keithw:  why are we even requiring a debhelper version in the first place?

andersk:
The debhelper version must support the compatibility level specified in `debian/compat`, which is currently 7. Additionally, the `override_dh_foo` syntax used in `debian/rules` requires debhelper 7.0.50.

If you want to kill the warning, bump `debian/compat` to 9 and the dependency to `debhelper (>= 9)`.

cgull:
I did more research and now I understand that a little better.  We're now using debhelper 9.

PR #765, opened 2016-06-09
Make OS X mDNS/Bonjour .local IPv6 addresses work.

cgull:
OS X advertises IPv6 link-local addresses for .local hostnames.
getaddrinfo() correctly reports interface identifiers for these, but we
were discarding them by using the socket peer host address instead.


cgull:
The problem this commit fixes:
`mosh -6 othermachine.local`
If mosh's hostname lookup happens to find the link-local address first, the client comes up but reports "mosh: sendto: No route to host" in the status bar on the top line.

PR #766, opened 2016-06-10
Add a little hint for first-time users

chenxiaoqino:
I was recommended to use mosh as "an in-place replacement for ssh" by my friends, and take a while to figure out the message until I realized I missed the line "The Mosh package should be installed on both the client and server."

If this line often corresponds to missing server-side installation, rather than other mysterious scenario, how about adding a hint?

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -420,7 +420,7 @@ if ( $pid == 0 ) { # child
     if ( $bad_udp_port_warning ) {
       die "$0: Server does not support UDP port range option.\n";
     }
-    die "$0: Did not find mosh server startup message. (Did you installed mosh on your server?)\n";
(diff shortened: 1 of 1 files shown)


scripts/mosh.pl
```diff
@@ -420,7 +420,7 @@ sub predict_check {
     if ( $bad_udp_port_warning ) {
       die "$0: Server does not support UDP port range option.\n";
     }
-    die "$0: Did not find mosh server startup message.\n";
+    die "$0: Did not find mosh server startup message. (Did you installed mosh on your server?)\n";
```
The-Compiler: installed -> install

chenxiaoqino: Sorry for the grammar mistake.

cgull:
I pulled this with slight grammar changes a while back.

PR #768, opened 2016-06-11
Add options to prefer v4 or v6 on connect; make prefer-inet the default.

### LINKED ISSUE #764: Mosh connections to older servers with both A and AAAA records fail with "mosh-server: Bad IP address" on newer client (1.2.5.95rc1)
Looks like the newer mosh clients (1.2.5.95rc1) fail connecting to older servers that are running on a dual-stack system with both A and AAAA records.  The failure looks as follows (with names and addresses changes):

```
bb@bb-mac:~$ mosh b@example.org
b@example.org's password:
mosh-server: Bad IP address (2001:DB8::)
Usage: mosh-server new [-s] [-v] [-i LOCALADDR] [-p PORT[:PORT2]] [-c COLORS] [-l NAME=VALUE] [-- COMMAND...]
Connection to example.org closed.
/usr/local/bin/mosh: Did not find mosh server startup message.
bb@bb-mac:~$
```

Note that example.org is available via both A and AAAA records, and a mosh directly to the IPv4 address works with no problems.

The mosh client install I'm using is on OS X 10.11.4 from the prebuilt package at  https://github.com/mobile-shell/mosh/releases/download/mosh-1.2.5.95rc1/mosh-1.2.5.95rc1.pkg.

I've repro'd the issue against mosh-server version 1.2.4a, on Ubuntu 14.04 LTS.

cgull:
This is an attempt to resolve #764 most of the time, while still preserving automatic connection to systems reachable via IPv6 only.
With this change, Mosh defaults to preferring IPv4 to IPv6 when both are available.  This means that dual-stack servers with IPv4-only `mosh-server` will still fail with the error seen in #764, but only when they are reachable only via IPv6.  I think it also means that Mosh will not get a lot of IPv6 use, since most servers with IPv6 will be dual-stack.  But I don't see a better way to handle this, if anybody else does please speak up.
On the plus side, I think this is also better for users connecting to dual-stack servers and then roaming or using VPNs, because IPv6 connectivity is a lot less universal than IPv4.

### DIFF
--- a/man/mosh.1
+++ b/man/mosh.1
@@ -143,23 +143,32 @@ Only use IPv4 for the SSH connection and Mosh session.
 .TP
 .B \-\-family=inet6
 Only use IPv6 for the SSH connection and Mosh session.  This and the
-following two dual-stack modes require Perl's IO::Socket::IP or
-IO::Socket::INET6 modules.
-
-.TP
-.B \-\-family=all
-Choose an address from all available IPv4 or IPv6 address, even for
-dual-stack hosts.  This is the most convenient option, but requires
-dual-stack connectivity when roaming with dual-stack servers.  This is
-the default.
+following modes require Perl's IO::Socket::IP or IO::Socket::INET6
+modules.
 
 .TP
 .B \-\-family=auto
 Autodetect IPv4 or IPv6 for hosts that only have addresses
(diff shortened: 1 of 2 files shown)


cgull:
Merged in af8ddd090e8a702063791ab3965a53b4f9818906.

PR #785, opened 2016-08-08
   Release 1.2.6, again

cgull:
This is the very same bits as #784, but from a new branch name in my GitHub fork instead of the `travis` branch, which I'd been reusing to run Travis on code before committing to mobile-shell/mosh.  Perhaps Travis doesn't actually like that.


glensc:
i've also using `travis` branch to test travis things and `push -f` on that branch. haven't got any issues using such behavior.

PR #789, opened 2016-08-15
Workaround lost MOSH CONNECT on FreeBSD 11

cgull:
This fixes an issue with Mosh on FreeBSD 11.0-BETA where the `MOSH CONNECT ...` message is lost in the pty driver.  This is a regression from earlier versions of FreeBSD and we don't see this issue anywhere else.  So I think there's a (minor) issue with FreeBSD ptys here.  See https://lists.freebsd.org/pipermail/freebsd-current/2016-August/062918.html for discussion.

However, the way `mosh-server` outputs messages to the pty is a bit sketchy and racy, so I've improved it here.


andersk:
This was committed as 4ad131a55d5a82a044a0c62428d6a488a97b46bf.

PR #790, opened 2016-08-15
Support OpenSSL's AES128_OCB

cgull:
This is an early implementation of support for the OCB implementation in OpenSSL 1.1.0.  Caveats:
- It depends on a pre-release version of OpenSSL.
- Only tested on FreeBSD.  (FreeBSD's multiple installs of OpenSSL can be quite annoying, but installing the "openssl-devel" package to get 1.1.0 was trivially simple.)
- Currently the code completely reinitializes OCB state for every message.

But it supports/passes test suites relevant for Mosh, and of course interoperates with the Rogaway/Krovetz implementation that Mosh normally uses.  Looking to the future, Mosh will probably gradually shift away from the Rogaway/Krovetz implementation, to external implementations, to get the benefits of ongoing maintenance, security analysis, etc.

This needs more work before it can go into Mosh; I want to at least see it support Linux and have better performance.  PRs gladly accepted :)  Also, there's no reason to pull it into master before there's some OpenSSL 1.1.0-release uptake by OS distributions.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -336,6 +336,33 @@ case "${with_crypto_library}" in
     ;;
 esac
 
+if test "${with_crypto_library}" = "openssl"; then
+  # push LDFLAGS/CFLAGS
+  old_LIBS="$LIBS"
+  old_CFLAGS="$CFLAGS"
+  # modify
+  LIBS="$LIBS $CRYPTO_LIBS"
+  CFLAGS="$CFLAGS $CRYPTO_CFLAGS"
+
+  AC_CHECK_DECL([EVP_aes_128_ocb],
+    [AC_DEFINE([HAVE_OPENSSL_OCB], [1],
+       [Define if the OCB AE algorithm is available in OpenSSL.])],
+    , [[#include <openssl/evp.h>]])
+
+  dnl AC_MSG_CHECKING([whether OCB is available in OpenSSL])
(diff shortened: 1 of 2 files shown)


eminence:
What's motivating this change?  Is the current impl considered old, or hard to maintain?

cgull:
The current impl has needed tweaks over time for various portability issues, `git log src/crypto/ocb.cc` for details, obviously.  I'm also a bit concerned that by not being in an established crypto library, it hasn't received general review and updates for implementation quality and side channel attacks and the like, since it was written in 2011 (or earlier?)  Weigh that against the significant point that we haven't seen security issues with that implementation yet.

cgull:
Also:  discussion on IRC has made me realize that right now, this code shouldn't autoconfigure itself into the Mosh build.  Instead it should be under a disabled-by-default option like `--enable-experimental-openssl-ocb` and left for crypto-nerds to discover and play with.

carloscabanero:
This is a good idea, just made a PR with more changes on the OCB code. Even Google is maintaining their own version for Android and isn't as well patched as mosh's.

cgull:
I just noticed this...if Google is maintaining a copy of the Krovetz/Rogaway OCB implementation, that'd be worth a look.  Where is it?

carloscabanero:
I think this was the one https://android.googlesource.com/platform/system/keymaster/+/master/ocb.c, I did an ediff when fixing the alignment and Mosh version seemed to be more on top. They have now added a different way to do the alignment themselves too.

PR #782, opened 2016-07-29
Style/perfomance/good practice

RussianBruteForce:
Just some improvements.

PR #800, opened 2016-08-29
Style/perfomance/good practice 2

RussianBruteForce:
Close #782 please.

For @cgull: use "Files changed" to make it easy to review.


andersk:
Because this includes a merge of #782, all of [cgull’s comments](https://github.com/mobile-shell/mosh/pull/782#issuecomment-237069559) about cleaning up the history still remain to be addressed. Each commit should be a single, logical, complete change. Things like typo fixes, formatting fixes, and compilation fixes should always be [squashed](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) into the commits that they fix. Clean history is important not just for reviewing the changes now, but also for later analysis (`git log`, `git blame`, `git bisect`, etc.).

configure.ac
```diff
@@ -115,7 +115,7 @@ AC_SUBST([HARDEN_CFLAGS])
 AC_SUBST([HARDEN_LDFLAGS])
 
 # Also check for a few non-hardening-related flags.
-MISC_CXXFLAGS=""
+MISC_CXXFLAGS="-std=c++11"
```
andersk: I don’t believe the project has made a decision about whether we are ready to move to C++11; support is still lacking on some of the older systems where Mosh may be used. But if we do, it should probably be turned on via the [Autoconf macro `AX_CXX_COMPILE_STDCXX_11`](https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_11.html), which supports multiple compilers.

eminence: FWIW, this PR fails to build on my fairly ancient moshing machine at work:

```
cc1plus: error: unrecognized command line option "-std=c++11"
```

I've for sure appreciated how portable mosh is to just about every linux environment I've been on.

RussianBruteForce: What compiler do you use? Is project built OK without this flag?

eminence: `c++ (GCC) 4.1.2 20080704 (Red Hat 4.1.2-48)`

If I revert a4c22e28767832f2cd16820f289b5d4616eeacff, the `-std=c++11` errors go away, but there are still problems in cc336c2908df8f1300b9c35fab381cc343608ebf.  If I also revert that, then this PR builds for me.

RussianBruteForce: RHEL5? Rly?

eminence: Yep, RHEL5.5.

RussianBruteForce: @eminence 
https://access.redhat.com/solutions/19458

eminence: I don't have control over this machine.  But sure, I could build a newer version of gcc, but I would very much rather not do that.  If the mosh team wants to drop support for these older compilers, that's fine as long as it is a deliberate decision.  I would hate to lose this support just because of an oversight (so I'm glad we've having this conversation here)

cgull: If it were 2017 already, I'd be comfortable with using C++11 :)

Right now, Ubuntu 12.04 and RHEL 5 are still supported releases, and neither have C++11 in their default compiler.  But most of C++11 was stable and available well before that, so it's a difficult call.  FreeBSD i386/amd64/arm/arm64 have moved to clang, OpenBSD is still using gcc 4.2, non-mainstream architectures tend to lag somewhat in their compiler support.  I lean towards saying "Not this year".  Then again, the next release might well be next year anyway.

It's not just compilers, a C++11 standard library and headers is also needed.  Another issue is that we'd need to get autoconf C++ detection/configuration working reliably across a variety of systems.

cgull:
@RussianBruteForce: I did try and clean this up myself a while back, spending quite a bit of time, and wasn't successful.  The merges were not hard to eliminate, but the complex changes to `calculate_timers()` and timestamp fetches turned out to be very difficult to untangle and review, and I did see at least one bug.  So I may yet extract the smaller changes from this PR, but I can't accept all of it as is.  A pull request really does have to be clear and understandable to be usable.

RussianBruteForce:
@cgull hi, I don't have any free time now, sorry about that. Maybe in 2017 I'll back for PR to Mosh, feel free to decline.

eminence:
This PR has been stalled for a while, there are conflicts, and the original submitted has said that they don't have time at the moment.  So I'm going to close this, but feel free to re-start the discussion (or open a new PR) later.  Thanks!

PR #802, opened 2016-09-01
Add links for Slackware mosh package

Strykar:
Slackware's package manager (pkgtool) does not track dependencies. Packages not bundled with the distribution have community maintained build scripts called $PKGNAME.slackbuild
These compile the package from source and create a Slackware package which pkgtool can install.

See https://slackbuilds.org/faq/#affiliation and http://www.linuxquestions.org/questions/slackware-14/slackware-14-2-is-coming-but-will-the-slackbuilds-will-also-be-updated-accordingly-4175575223/#post5517961

### DIFF
--- a/slackware/Slackware.README
+++ b/slackware/Slackware.README
@@ -1,0 +1,10 @@
+REQIURES:
+python-dateutil
+python-gflags
+protobuf
+perl-IO-Tty
+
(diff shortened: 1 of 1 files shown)


Strykar:
Should've been sent to the website repo

PR #808, opened 2016-10-02
Use a table lookup for base64 decode

cgull:
This replaces the linear lookup that I originally wrote, which I've long felt a little uneasy about.  My original reason to do it that way is weak, and it seems better to use a table lookup as other crypto base64 implementations use, even if I don't actually see any performance or security issue here.


cgull:
Merged as 9005ea6cf39ac77b8edfabb471b05207c5fd50a6

PR #815, opened 2016-10-29
Spin fix

cgull:
This fix was a long time in coming.

This fixes a problem I discovered where `mosh-server` would get into a state where it would call select() with a timeout of 0, and could spin for up to 3 seconds (until the next ack was sent to the client).  This happens when the client writes to the terminal in some way that leaves the display unchanged (such as `"x\b"`).  It took me...quite a while to discover that there was another very similar bug where input prediction's rapid acks could also cause a spin for 1 millisecond.

The fix for the second problem was simple, but the theoretically best fix for the first problem (making `Terminal::Complete::operator==()` do its comparison with the correct depth, and always agree with `Terminal::Complete::diff_from()`) turns out to be a large project, with implications for things like object lifetime and performance.  I might finish that half-done work, but it doesn't make sense for a near-term release.  So instead I have this fix that simply adjusts the timers involved to avoid the zero timeout.  (There's also a band-aid to convert zero timeouts to 1ms timeouts, which will reduce the impact of anything like this in the future.)

@keithw, I believe that fix will do something fairly reasonable, but would you have a look at 
https://github.com/mobile-shell/mosh/commit/34e69867750578b4cc6a1c137113913a30107d4b for me and see if you agree?

In this PR, the fixes are after the tests, to make it easy to run the tests on the buggy code.  When I merge this into mobile-shell/mosh, I'll reorder the commits so the fixes are before the tests, so as to not break future bisecting.

### DIFF
--- a/src/tests/network-no-diff.test
+++ b/src/tests/network-no-diff.test
@@ -1,0 +1,60 @@
+#!/bin/sh
+
+#
+# This is a regression test to ensure that Mosh does not spin
+# on updates that do not actually change the framebuffer.
+#
+
+. $(dirname $0)/e2e-test-subrs
+PATH=$PATH:.:$srcdir
+# Top-level wrapper.
+if [ $# -eq 0 ]; then
+    e2e-test $0 baseline post
+    exit
+fi
+
+# OK, we have arguments, we're one of the test hooks.
+if [ $# -ne 1 ]; then
+    fail "bad arguments %s\n" "$@"
+fi
+
+baseline()
+{
+    # Generate updates that don't change the screen
+    local i
+    i=0
+    while [ $i -lt 10 ]; do
+	printf 'x\b'
+	i=$((i + 1))
+	sleep 1
+    done
+}
+
+post()
+{
+    # Extract server run time.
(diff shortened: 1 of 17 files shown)


src/tests/network-no-diff.test
```diff
@@ -36,7 +36,7 @@ post()
     runtime=$(sed -E -n 's/.*@@@ runtime: (.*) @@@.*/\1/p' "$(basename $0).d/baseline.tmux.log")
 
     # If this system can't actually report runtime, bail.
-    if [ "$runtime" -eq "-" ]; then
+    if [ -z "$runtime" -o "$runtime" -eq "-" ]; then
```
andersk: Use `[ -z "$runtime" ] || [ "$runtime" -eq "-" ]` instead. The `-a` and `-o` operators are less portable and also extend the grammar ambiguously such that they won’t always do what you mean even on systems that support them.
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html#tag_20_128_16

cgull: Done-- not a big issue for test/build code, but still worth fixing here, in your shell-test PR, and the new tests I just committed.

cgull:
Merged in 52d10cf21105e1fa831431c48ecb47495be2a7f7

PR #818, opened 2016-10-30
Fix most shell hygiene issues found by shellcheck

andersk:


cgull:
For the record, I also ran `checkbashisms`, and edited the scripts to use `posh` and ran `make check` as well.  No issues found, except that `posh` seems to have an issue where children in a pipeline don't get `SIGPIPE` when writing to a downstream that has exited.

cgull:
Merged in b76d1bd526f9ab62d27d916307549040bdf7af7b

cgull:
And I beleive I've fixed the `posh` issue in 4bc9e5c6ab5f9c0980f0dbd6606380198f13277b.

Issue #812, opened 2016-10-13
do OS X package build with Travis


cgull:
Currently OS X package builds are done by hand.  Now that we've got Travis, it should be pretty easy to do them there for better visibility and reproducibility.  The trickiest bit is probably getting Travis to do the package build _only_ for releases (probably only on tags).

PR #822, opened 2016-11-06
Travis osx package build

cgull:
This is for issue #812, following discussion on the mosh-devel mailing list.
This gets us an automated build with hopefully-good reporting of the components that go into it (OS X, Xcode, Homebrew, Mosh Git).  It creates a package, and a separate tarball reporting the build environment and configuration.

I've given up on deterministic/reproducible builds for the moment.  Apparently Apple's toolchain (in particular the linker) is incapable of doing deterministic builds.  It appears the bitcoin crowd has managed to get deterministic builds, but only by cobbling together a Linux VM environment, cross-compile tools (some of which are fairly broken), and Apple SDK components.  Not quite ready to go there yet.

I've mostly cleanly separated the Travis build from the package manager install from the package build itself.  From this point it may not be too hard to compose a more independently trustable build VM with either manual builds, or using Gitlab as a conduit to automatically start builds.

See https://github.com/cgull/mosh/releases/tag/mosh-1.2.6.98.8 and https://travis-ci.org/cgull/mosh/jobs/173651031 for the latest package build using this setup.  Pay no attention to the release number here, it hasn't got any relation to actual Mosh releases and shouldn't be viewed as an early release candidate.  The Travis package build is activated by pushing a tag, and I needed to name the tag *something*.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ addons:
     - perl
 
 before_install:
-  # Get full repo and describe this
+  # Get full repo for `git describe`
   - git fetch --tags --unshallow
   # OS X dependencies.  The package_deps step takes 15 minutes or so
   # and is fairly quiet, so wrap it in travis_wait to keep it from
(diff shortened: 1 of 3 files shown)


cgull:
Pulled, but I need to finish connecting it to `mobile-shell/mosh`, which I'll do at release time.

PR #823, opened 2016-11-08
Make all commands properly support --help and --version

andersk:
These should output to stdout and exit with status 0.  Passing `std-options` to `AM_INIT_AUTOMAKE` causes `make installcheck` (hence also `make distcheck`) to verify this.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@
 
 AC_PREREQ([2.61])
 AC_INIT([mosh], [1.2.6], [mosh-devel@mit.edu])
-AM_INIT_AUTOMAKE([foreign -Wall -Werror])
+AM_INIT_AUTOMAKE([foreign std-options -Wall -Werror])
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 AC_CONFIG_SRCDIR([src/frontend/mosh-client.cc])
 AC_CONFIG_MACRO_DIR([m4])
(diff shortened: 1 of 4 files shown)


cgull:
* Could we also support `-h` and `-?` consistently?
* Comment: this will conflict with the format strings reformatting in #826.

andersk:
> Could we also support `-h` and `-?` consistently?

Suggestion or request? The point of this commit is specifically to satisfy [`std-options`](https://www.gnu.org/software/automake/manual/html_node/List-of-Automake-options.html#index-std_002doptions) (and, more or less by accident, the [GNU coding standards](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html)) so that `make installcheck` has something to check. I’m a little squeamish about making unneeded changes to the `mosh-server` interface, since it’s not really user-facing and changes might make future feature detection code more fragile.

cgull:
Sorry, I should have been clearer.
* Supporting `-h` and `-?` is a suggestion for UX; these are traditional help flags that users may try in attempting to learn what a program is or does.  Like you, I'm a bit squeamish about altering command parsing in `mosh-server`, but these are simple adds and seem low risk.
* Making all the programs provide the same subset of `--help` `-h` `-?` is a request-- we should at least be self-consistent on this.

Also, a topic for discussion:  in mosh-client.cc you have added ad-hoc code for doing `--help` and `--version` long options.  Ad-hoc getopt-related code always bothers me because it almost always grows; maybe it's time to switch to `getopt_long()`.  The GNU implementation is of course widespread; the BSDs and OS X have 95% compatible reimplementations available.  Casual Googling shows that even Solaris and AIX have had it available for a decade.  How do folks feel about this?

There's also ad-hoc code in `mosh-server` that you've added to, but since I am rather squeamish about reworking that to use `getopt_long()`, I'm OK with that.

andersk:
This was committed a while ago as 94027efc455855d4ed777048ad2e8fef73ef02d8.

PR #824, opened 2016-11-08
Add a trivial test that the mosh script works without needing tmux

andersk:
None of the previous tests even ran the `mosh` script unless tmux ≥ 1.6 is installed.  Thus `make check` was “passing” on, e.g., RHEL 6 even though its Perl is too old (5.10.1).

### DIFF
--- a/src/tests/inpty.cc
+++ b/src/tests/inpty.cc
@@ -1,0 +1,104 @@
+/*
+    Mosh: the mobile shell
+    Copyright 2012 Keith Winstein
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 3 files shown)


andersk:
This needs more work to avoid failing when stdin isn’t a tty.

src/tests/local.test
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -eu
+out=$(./inpty ../../scripts/mosh --client="../frontend/mosh-client" --server="$PWD/../frontend/mosh-server" --local --bind-server=127.0.0.1 127.0.0.1 -- printf '\150\145\154\154\157\n')
```
cgull: Something like `printf 'he%s\n' lo` would be more readable.

src/tests/inpty.cc
```diff
@@ -0,0 +1,104 @@
+/*
```
cgull: Might it be better to do this in Perl?  It'd be much simpler, and independent from configure and Mosh utility code.

andersk: Then we’d need IO::Pty again.

cgull:
I decided this was good enough for now and took it anyway.  I did indeed run into configuration issues on FreeBSD, and I also made a small correction to your commit message (`s/1.6/1/8/`).

I also tried mosh-1.2.2, which relies on `IO::Pty`, on Cygwin 2.6.0 (which is not quite current).  It worked fine.

src/tests/inpty.cc
cgull: I decided to take this, deciding better was the enemy of good enough.

I also tried mosh-1.2.2 on Cygwin 2.6.0 (not quite current) and it worked fine.  What other pain was `IO::Pty` causing?

cgull:
Argh.  Stale pages in browser.

PR #825, opened 2016-11-10
Restore Perl 5.10 support

andersk:
Apparently a RHEL 6 user [complained](https://bugzilla.redhat.com/show_bug.cgi?id=1393100) that Mosh needs Perl 5.14 now. I don’t know if we care, but in case we do, here’s a commit restoring that support with the help of Socket::GetAddrInfo from CPAN (packaged in EPEL 6 as perl-Socket-GetAddrInfo).

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -30,18 +30,24 @@
 #   this exception statement from all source files in the program, then
 #   also delete it here.
 
-use 5.14.0;
+use 5.10.0;
 
 use warnings;
 use strict;
 use Getopt::Long;
 use IO::Socket;
 use Text::ParseWords;
-use Socket qw( getaddrinfo getnameinfo AI_CANONNAME AI_NUMERICHOST NI_NUMERICHOST
- IPPROTO_IP IPPROTO_IPV6 IPPROTO_TCP IPPROTO_UDP );
+use Socket qw( IPPROTO_IP IPPROTO_IPV6 IPPROTO_TCP IPPROTO_UDP );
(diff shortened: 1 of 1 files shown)


cgull:
This looks like it's intended to work on all Perl versions >= 5.10, but it fails Travis and OS X `make check`.

andersk:
Whoops, I’d left in some intentionally broken debugging residue; fixed.

cgull:
Pulled, thanks!  Not working through this issue was a gap in my original work in 389fc232051087f2430e9e7b05f770a38069cb39.  It works fine on my OpenBSD 5.2 system with Perl 5.12.2.

For the record:  This isn't explicit support for systems this old-- we can only really support OSes that are themselves still being supported.  But changes which improve backward compatibility without affecting current OSes, like this one, will definitely be considered.

achernya:
I've inspected this patch for EL5 and EL6 -- it looks like we can actually support perl 5.8.8 as-is, if we only import IPPROTO_TCP and not the other variants.

jennakwon06:
I am having this issue. What's the workaround?

andersk:
@jennakwon06 Install Mosh from EPEL, which has the patch backported; or compile it from current Git; or use the 1.3.0 release candidates.

andersk:
@jennakwon06 My mistake: apparently the EPEL update with this patch only transitioned to EPEL 6 stable today and is still in EPEL 5 testing. See the [EPEL bug](https://bugzilla.redhat.com/show_bug.cgi?id=1393100).

PR #829, opened 2016-11-13
debian: Build-Depend tmux for better test coverage

andersk:


andersk:
This will probably just make Debian/Ubuntu builds fail with the same `/dev/ptmx` problem.

cgull:
Once Debian fixes their buildroot, we can still use this.

eminence:
`tmux` was added as as `Build-Depends` in 0fc950bd8a3469bfb19c3049d35480a20205e66e so i think this PR can be closed

PR #830, opened 2016-11-17
Copy/paste fix

cgull:
Well, that took a while.

My earlier performance commits changed how the wrapped-line bit is stored, and that change caused Mosh to not clear the wrap bit when a long line was replaced by a short one on screen.  So I essentially reverted that change.  But that broke one of the regression tests in a subtle way.  Fixing that took even longer.

I believe this to be well and truly fixed, though I have not had a lot of interactive usage yet.  If you have this problem, please try this code.


cgull:
I still haven't had much interactive use ($work just has not gone that way this last couple of days), but I did some testing of various terminal emulators and all looks good.

Issue #832, opened 2016-11-30
Long delays or stalls when refreshing terminals with lots of static text


karlkec:
There are sometimes excessively long refresh delays with a slow link and a terminal with lots of unchanging text.  When switching between tmux windows with lots of static text on a 475x125 character console, it can take many seconds to hours to refresh the terminal. Sometimes it completely stalls with the terminal partially refreshed, and while refreshing it saturates the server-to-client connection. Neither the client nor server are taking excessive CPU. The client indicates a lost connection, but if I type a hot key on the client to switch tmux at the far end to a sparsely populated window, the whole thing recovers. It fails again as soon as tmux is switched to a window with lots of text.
An ssh connection at the same time to the same tmux window refreshes in a couple of seconds.

andersk:
These are the symptoms of a MTU problem with your network. Mosh clears the Don’t Fragment flag in its packets, so large packets are supposed to get fragmented by the network and reassembled by the recipient, but some misconfigured networks drop these packets on the floor instead.

If you aren’t already using Mosh 1.2.6 on the server, that may help as it reduced the default MTU from 1380 to 1280 (#773). If you’d like to help us figure out how to work around these broken networks, see [this comment](https://github.com/mobile-shell/mosh/issues/299#issuecomment-231940226).

cgull:
Anders is correct in suggesting that this is probably an MTU problem, but there is one other possibility worth mentioning:  Mosh communicates via application datagrams that can theoretically reach 50MB in size.  These are fragmented into separate UDP datagrams by the sender and reassembled by the receiver.  If any single fragment goes missing then the entire message is lost.  (UDP fragmentation may also happen underneath this layer as well.)  Also, the fragment reassembly is quite simple-minded, and if fragments for two separate messages are interleaved, one of the messages will be lost.

I've not actually seen this problem, but your console windows are pretty big (larger than anything I regularly use), and I think messages for windows this big might be at some risk of never being successfully delivered on networks with high loss, or with queues too short to handle all the packets in a message.  Can you characterize your network link's performance at all?

cgull:
If your network link is really, really low-bandwidth, I think maybe it's also possible that Mosh's timeout constants might cause it to send data at a higher rate than the link can actually deliver.  ("Low-bandwidth" would be something like ISDN-B (128kb/s) or slower.)

karlkec:
[...]

Issue #747, opened 2016-04-28
mosh-server can generate high traffic if a client can't hear it


cgull:
Tonight's discovery:
If:
* a server is has an active application running (a tmux session that is being used from another client, for example) and the session generates significant changes
* the server can hear the client but the client does not hear the server

then the server's message traffic to the client can grow somewhat high-- a bandwidth graph I was looking at showed spikes to 1Mb/s of traffic for each update, whatever that means.

Given mosh's design, this isn't unexpected, but seems a little suboptimal for pay-by-use networks.  I saw this while doing a router/server upgrade and fixed the routing problem, but this might be worth a little time to review/improve.

eminence:
If the client could tell the server "I haven't heard from you in `X` seconds", would that let the server do something intelligent?

Issue #299, opened 2012-07-16
Support smaller MTU when PMTUD is not available (e.g., some VPNs)


joehorsnell:
Hi,

I cannot get my mosh session to reconnect over IPsec VPN.

Server:
- Linux VM (Centos 6.2, kernel 2.6.32-220.13.1.el6.x86_64) in ESX cluster
- mosh 1.2.2

Client:
- Mac OS X 10.7.4 (Darwin Kernel Version 11.4.0)
- mosh 1.2.2

If I start a mosh session to the server over VPN, then disconnect the VPN, I get the blue banner at the top of my session, "mosh: Last contact x seconds ago" but if I reconnect the VPN the session does not reconnect. I just have to quit.

Any ideas? Would you expect this to work over VPN?

Thanks in advance,

Joe.

AngerM:
I'm also running into this issue using 1.2.5 w/ OS X client and Centos Server. I'm 99% sure this is an MTU issue. I notice it most often when either reconnecting or when a lot of things on the screen change then I get 'disconnected' from the server, but the server is still receiving all my commands (this can easily be seen using tmux and connecting to the same tmux session via ssh and mosh).

cgull:
There are clearly some problems with Mosh and VPN and public networks, as we see here, on #773, and IRC.  But reports of "It doesn't work for me" aren't enough for us to be able to clearly determine what's going on.

If you run into problems with VPN, it will help us greatly if you can:
* give us a packet capture of a session that hangs, preferably at both client and server
* figure out what the path MTU is from your server to your client, manually with `ping` or more automatically with `tracepath` or similar
* name the public network and the VPN software you're using

I know these aren't always easy asks, technically or politically, but they will help us better establish what's going on with VPNs.  Thanks!

jtlawson:
At the university where I work, we have a Cisco VPN product.  When connected to the VPN, my MTU is 1100 bytes.  (ping -s 1072 to the mosh server works, ping -s 1073 and above doesn't.)  I experience the same behaviors described by others.

I was running mosh 1.2.6 on both client (OS X 10.11.6, mosh from macports) and server (RHEL 6, RPM built from source.)

I patched in PR #836 on the server but it did not seem to make a difference in behavior.

Then I modified src/network/network.h and dropped DEFAULT_IPV4_MTU to 1072, rebuilt the server, and since that time I haven't had troubles when connected to the VPN.

This should work for me until mosh performs some sort of ICMP-less PMTUD.

[...]

PR #836, opened 2016-12-22
Exponential backoff to prevent aggressive resends

cgull:
If a client has one-way connectivity to a server such that it keeps requesting the same state from the server, the server may spam the client with updates on its interactivity-timeout interval of 40ms until 10s have gone by, at which point it reverts to a 3s interval.  With large terminal windows, this can cause excessive bandwidth usage.  If the link has limited bandwidth causing it to drop some packets from a large message, the flood of these packets can also prevent the client from receiving a new state successfully.  See #747 and #832.  I also think this issue might be responsible for some of the "VPN" issues that we have lumped under #299.

This implements a rather crude exponential backoff for resends, simply delaying the next send by the amount of time since the first unsuccessfully sent state.  The actual delays calculated are rather noisy and sometimes larger than they need to be.  The backoff is initiated after at least two messages have been sent to the remote and not acked; we need to work from the timestamp of the first unsuccessful send, not the timestamp of the last successful send (which may have been unusefully long ago), and one of the messages sent in this time may have been an ack-only message, which doesn't count.  This inherently adds a bit of conservatism to the backoff.  The backoff is limited by the existing calculations for `next_send_timer` and `next_ack_timer` which currently impose a ceiling of 3 seconds.

I was concerned about the impact of exponential backoff-- I thought that its much less frequent sends on lossy networks would result in laggier interactive behavior.  But in testing with Network Simulator on OS X with 50% packet loss and 1s+ latency, I don't see any change in behavior-- Mosh seems fine.

Comments and testing greatly appreciated.

### DIFF
--- a/src/network/networktransport-impl.h
+++ b/src/network/networktransport-impl.h
@@ -81,6 +81,7 @@ void Transport<MyState, RemoteState>::recv( void )
       throw NetworkException( "mosh protocol version mismatch", 0 );
     }
 
+    uint64_t now = timestamp();
     sender.process_acknowledgment_through( inst.ack_num() );
 
     /* inform network layer of roundtrip (end-to-end-to-end) connectivity */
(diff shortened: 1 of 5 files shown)


cgull:
Testing this code again, I don't see it working reliably on a link with no delay and 50% loss.  Mosh still is too aggressive about resends but this isn't the fix for it.

PR #837, opened 2016-12-22
Release 1.3.0 candidate

### LINKED ISSUE #654: Please place 1.2.5 OS X packages as GitHub releases
We have an automated mechanism that downloads new releases from this Github repo.  It's been working great so far, but your OS X packages are added separately, not as Github releases, and so our mechanism didn't see them.  It would make me quite happy if you provided them as releases instead of custom links.  Thanks! :)

cgull:
This is my current plan for 1.3.0 release.

This will add some significant changes to Mosh's release process:

* Semantic versioning
* The Travis OS X build generates a distribution tarball and OS X package, and uploads them to Github.  It also generates copious logging-- it's not a fully reproducible build (both Travis and OS X make this quite challenging) but it is intended to make the build more repeatable and traceable.

Note, this is not 100% complete, `travis.yml` needs to be switched from `cgull/mosh` to `mobile-shell/mosh`, and the `mobile-shell/mosh` project and its Travis config need to add a key to allow Travis to upload the artifacts.

@keithw, are you OK with these changes?  We'd be switching from generating these artifacts manually and uploading directly to mosh.org, to taking Travis artifacts from Github and copying them to mosh.org.  One upside is that this would resolve #654.  One downside is that if Travis is having problems then we can't generate a release this way (and Travis has been having lots of problems lately).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,17 +36,14 @@ before_install:
   # and is fairly quiet, so wrap it in travis_wait to keep it from
   # being killed.
   - |
-    (
-      set -e
-      if test osx = "${TRAVIS_OS_NAME}"; then
-        macosx/brew-deps.sh install
+    if test osx = "${TRAVIS_OS_NAME}"; then
+      macosx/brew-deps.sh install &&
         if test -n "${TRAVIS_TAG}"; then
           travis_wait 30 macosx/brew-deps.sh package_deps
         else
           macosx/brew-deps.sh deps
         fi
-      fi
-    )
+    fi
 
 # Use before_script to report the build configuration.
 before_script:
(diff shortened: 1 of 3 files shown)


cgull:
https://github.com/cgull/mosh/commits/mosh-1.3.0-rc0 and https://github.com/cgull/mosh/releases/tag/mosh-1.3.0-rc0 are a sample build linking to Travis and showing the GitHub release and artifacts.

keithw:
This seems great for the OS X binary .pkg.

For the official source release, though, I'm not seeing as much value in going through Travis -- this puts us in the position of PGP-signing the hash of a .tar.gz that was created on a Travis VM, which makes me mildly uncomfortable...

cgull:
Comparing that example Travis distribution tarball and one generated on a current Linux system, the changes are confined to autoconf/automake generated files (as they should be).  For this example, most of it's a boilerplate addition of `runstatedir` but there's a hunk of code in `install-sh` testing `mkdir` that changes how it handles its temporary directory.  Would you be comfortable doing a comparison like this against your trusted machine and then signing the Travis-generated tarball if its content looks OK?

keithw:
If we can script that comparison, sure, I guess that would work. Ideally it would be so reproducible that you wouldn't be able to tell the difference between signing the Travis-built source tarball and one built on any other Linux machine. :-) (Are these differences just because of differing automake/autoconf versions between the Travis VM and your Linux system?)

cgull:
Not scriptable, no-- requires manual review from somebody.

Differing versions seem to be the cause of it, yes, but I'm reluctant to rule out the possibility that the same autoconf/automake versions run on different OSes might produce different output.

The reason I'd like to do this is that currently, you do `make dist` on some private machine of yours, sign that, and put it up on mosh.org as a release.  I trust you just fine with that :) but I think it would be nicer to expose that publicly as part of our release process.  (It does mean trusting Travis OS X images, which are a juicy target for many actors.)  I'm OK with not changing this though.

I'm currently looking at the OS X `clock_gettime()` issue again which may end up touching this anyway.

cgull:
Travis wraps shell commands and configuration from `.travis.yml` into a single shell script and uses advanced bash features to manage execution.  They mention intercepting/manipulating `$?`, for example. The old code broke when I switched Travis images at some point, I think.  I removed `set -e` to avoid script exits.

PR #838, opened 2016-12-24
Detect availability of clock sources at runtime

### LINKED ISSUE #807: mosh fails with _clock_gettime error on OS X 10.11.6 and Xcode 8
mosh installed through the brew package manager runs, but when I try to connect to a server with an SSH key, it fails with "Symbol not found: _clock_gettime". The binary from https://mosh.org/#getting or building from source does not have this problem.

```
$ brew config

HOMEBREW_VERSION: 1.0.4-7-g6e57355
ORIGIN: https://github.com/Homebrew/brew.git
HEAD: 6e5735565b9a31ec2b01e02a479a361189729c22
Last commit: 5 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 109939afe4af2dd8742cbc31af1bc2b0e04a6556
Core tap last commit: 27 minutes ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 8.0 build 800
Git: 2.10.0 => /usr/local/bin/git
Perl: /usr/bin/perl
Python: /usr/local/var/pyenv/shims/python => /usr/local/var/pyenv/versions/2.7.12/bin/python2.7
Ruby: /usr/local/var/rbenv/shims/ruby => /usr/local/var/rbenv/versions/2.3.1/bin/ruby
Java: 1.8.0_66, 1.7.0_45
macOS: 10.11.6-x86_64
Xcode: 8.0
CLT: 7.3.1.0.1.1461711523
X11: 2.7.9 => /opt/X11
```

```
$ mosh --version

mosh 1.2.6 [build mosh 1.2.6]
Copyright 2012 Keith Winstein <mosh-devel@mit.edu>
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

```
$ mosh --ssh="ssh -i /Users/user/.ssh/id_rsa" root@1.2.3.4

dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/mobile-shell/1.2.6_2/bin/mosh-client
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/mobile-shell/1.2.6_2/bin/mosh-client
  Expected in: /usr/lib/libSystem.B.dylib

Trace/BPT trap: 5
```

This was originally reported at https://github.com/Homebrew/homebrew-core/issues/5220 and patched at https://github.com/Homebrew/homebrew-core/pull/5247, but it's an upstream issue, so I'm reporting it here.

The fix in the brew formula was to set `ENV["ac_cv_search_clock_gettime"] = "no"`.

cgull:
`clock_gettime()` was introduced in OS X 10.12 and iOS 10.  However, an executable built with SDKs for these OS versions will not have that function available when run on an older OS version and will crash at runtime.  Try all available clocks at runtime, detect `clock_gettime()`'s presence for Apple OSes, and improve error handling for the Mach clock a bit.

This has been tested for all clocks, but has not yet been tested for the crashing situation described above, or for iOS.  I expect this to be a no-op for our 1.3.0 OS X package build since it will be built for an older OS X version.

Fixes #807, or at least gets one step further toward fixing it.

### DIFF
--- a/src/util/timestamp.cc
+++ b/src/util/timestamp.cc
@@ -37,12 +37,15 @@
 #include <errno.h>
 
 #if HAVE_CLOCK_GETTIME
- #include <time.h>
-#elif HAVE_MACH_ABSOLUTE_TIME
- #include <mach/mach_time.h>
-#elif HAVE_GETTIMEOFDAY
- #include <sys/time.h>
- #include <stdio.h>
+#include <time.h>
+#endif
+#if HAVE_MACH_ABSOLUTE_TIME
+#include <mach/error.h>
+#include <mach/mach_time.h>
+#endif
+#if HAVE_GETTIMEOFDAY
+#include <sys/time.h>
+#include <stdio.h>
 #endif
 
 static uint64_t millis_cache = -1;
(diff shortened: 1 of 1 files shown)


cgull:
@carloscabanero, can you try this in your environment?

carloscabanero:
Ack, traveling now but give me a couple hours.

carloscabanero:
Sorry for not getting back to you before. This is a good approach, and I've been thoroughly testing it on all devices with success, and checking on debugger the path that it is taking.

I found one thing though. As we are cross compiling for different iOS or OSX versions, if you compile on < 10.12, to use on >=10.12, then HAVE_CLOCK_GETTIME flag will be false on the configure test, and you will always go the MACH route. Just removing the #if HAVE_CLOCK_GETTIME should work fine, but I guess that may cause problems in other platforms when compiling.

cgull:
This should be resolved now in commit b2eac32ebc580f75e51a1608c61657ce24f027bc, and follow-up commit ba5f33fd764b58b5282e46f62866968be88dfdb9 for non-Apple OS compatibility.

PR #839, opened 2016-12-27
fix: stop overwriting command from `--server`

Blaok:
Currently the options specified with `--server=COMMAND` got overwritten by some built-in parameters in the script, which could cause problems when I intend to do
```
mosh example.com --server='mosh-server new -i :: -l LANG=en_US.UTF-8'
```
while getting something like
```
mosh-server new -i :: -l LANG=en_US.UTF-8 new -c 256 -s -l LANG=en_US.UTF-8
```
on the server side. The latter will bind to a wrong IP in some of my multi-homed servers.

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -382,6 +382,11 @@ if ( $pid == 0 ) { # child
   if ( scalar @command > 0 ) {
     push @server, '--', @command;
   }
+  
+  if ( $server cmp 'mosh-server' ) {
+    undef @server;
+    my @server = ( '' );
+  }
 
   if ( defined( $localhost )) {
     delete $ENV{ 'SSH_CONNECTION' };


scripts/mosh.pl
```diff
@@ -382,6 +382,11 @@ sub predict_check {
   if ( scalar @command > 0 ) {
     push @server, '--', @command;
   }
+  
+  if ( $server cmp 'mosh-server' ) {
```
cgull: Unfortunately, this will break the existing, common usage of `mosh  --server=/usr/local/bin/mosh-server`, so it isn't the right fix.  A workaround is to set up a `mosh-server` wrapper script on your server, but we should find a better fix for this problem.

cgull:
Also, look at the `--bind-server` option, which is intended to help with this type of problem.

scripts/mosh.pl
Blaok: You are right I could having been using a wrapper on the server side. Problem solved! Thanks.

PR #841, opened 2016-12-27
Memory Alignment issues on ARM processors

carloscabanero:
Unaligned data on ARM architectures do not perform efficiently unaligned
memory access, and in the case of ARMv7 and iOS it completely breaks. The OCB
algorithm dereferences a uint64x2_t pointer, and is replaced by a
memcpy to avoid penalties when trying to align it.

More info https://brewx.qualcomm.com/bws/content/gi/common/appseng/en/knowledgebase/docs/kb95.html

### DIFF
--- a/src/crypto/ocb.cc
+++ b/src/crypto/ocb.cc
@@ -271,8 +271,9 @@
 	static block gen_offset(uint64_t KtopStr[3], unsigned bot) {
 		const union { unsigned x; unsigned char endian; } little = { 1 };
 		const int64x2_t k64 = {-64,-64};
-		uint64x2_t hi = *(uint64x2_t *)(KtopStr+0);   /* hi = A B */
-		uint64x2_t lo = *(uint64x2_t *)(KtopStr+1);   /* hi = B C */
+		uint64x2_t hi, lo;
+ 		memcpy(&hi, KtopStr, sizeof(hi));
+ 		memcpy(&lo, KtopStr+1, sizeof(lo));
 		int64x2_t ls = vdupq_n_s64(bot);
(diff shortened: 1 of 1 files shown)


cgull:
Just so I understand here:  what is at issue here?  Performance, correctness, or exceptions killing Blink?

And I assume this problem happens all the time, because the old code always does a misaligned move (assuming that `uint64x2_t` is a 16-byte-aligned type).  Is that correct?  And are you encountering this on 32-bit and/or 64-bit?

carloscabanero:
This seems to be a problem that affects all ARM architectures, but the problem scope varies depending on MMU and OS. 64 bit devices fix it but incur in additional cost to do so. I found it on Blink on iOS 32bit devices where it crashes Blink, but given Qualcomm description I guess on other Android systems might crash too. Others like OpenSSL or libgit have had the same problem before.

The solution not only fixes the crash but according to them it also improves performance as the MMU doesn't have to trap the misalignment and fix it, saving 10-12 CPU cycles.

I have applied this patch to Blink and it is working perfectly on iPad3, iPhone 5, etc...

Sorry about the brief explanation, I will do a better job in the next PRs.

fornwall:
It seems that the fix was committed in https://github.com/mobile-shell/mosh/commit/0ceb4f26f4bf10656ea55f72e2388159c761d737 and this PR can be closed, right?

cgull:
Yes, it should be fixed, though I've not directly confirmed with an armv7 or armv8 build.
I may yet adopt Google's fix which uses a NEON fetch instruction, but this certainly works for now.

PR #842, opened 2016-12-27
Renew socket when closed

carloscabanero:
The Socket might be renewed during the lifetime of the Mosh session, but
if it was closed for some reason, the client hangs.
We have added support to check the state of the socket, and in case it
has been closed externally (systems with low resources), to reopen it and continue.

### DIFF
--- a/src/network/network.cc
+++ b/src/network/network.cc
@@ -442,7 +442,12 @@ string Connection::recv( void )
 	   || (e.the_errno == EWOULDBLOCK) ) {
 	assert( !islast );
 	continue;
-      } else {
+      } else if (e.the_errno == ENOTCONN) {
+	hop_port();
+	socks.erase(it--);
+	continue;
+      }
+      else {
 	throw;
       }
     }


src/network/network.cc
```diff
@@ -442,7 +442,12 @@ string Connection::recv( void )
 	   || (e.the_errno == EWOULDBLOCK) ) {
 	assert( !islast );
 	continue;
-      } else {
+      } else if (e.the_errno == ENOTCONN) {
```
cgull: Several problems here:
* This fails Travis CI on Linux
* This will break servers, if they ever receive `ENOTCONN`-- the server only has one socket, and `hop_port()` is not appropriate
* This uses the iterator `it` after it's been invalidated by `.erase()`
* If the failing socket is the first one in the container, `it--` moves the iterator past the beginning of the container

cgull:
A bit of googling suggested that this is an iOS specific problem where suspended apps start getting `ENOTCONN` when they are resumed in some circumstances.  Can you describe the issue you're seeing?

achernya:
Closing due to lack of activity.

PR #851, opened 2017-01-31
Fixes for GCC 7

andersk:
We’ll need something like this to build with GCC 7 and `--enable-compile-warnings=error`. (A [Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=853558) about this was reported by the Debian GCC maintainers.)


andersk:
Replaced with #852.

PR #852, opened 2017-02-01
Handle GCC 7's new -Wimplicit-fallthrough.

cgull:
This is an alternate fix to the one on #851.  Syntax checked on an online C++ playground with GCC 7, but not actually tested on Mosh + GCC 7.  @andersk, do you have any strong preference?

Curiously, clang++ 4.0 on FreeBSD seems to have the option but doesn't seem to actually warn.


cgull:
Turns out the other GCC 7 issue is in dead code.

andersk:
Yeah, this looks better and works for me with GCC 7.

PR #857, opened 2017-02-22
Ensure a UTF-8 locale is available for tests

cgull:
I think this will fix the Debian build failures we're experiencing, but cannot fully test this, not being the Debian package maintainer :)

I think it'll work on other platforms too, though I have a perhaps-confused memory of some platform that does not use `en_US.UTF-8` but instead some different naming convention.  (On some platforms, `locale -a` will output `en_US.utf8` even though the other form is required.)

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: mosh
 Section: net
 Priority: optional
 Maintainer: Keith Winstein <keithw@mit.edu>
-Build-Depends: debhelper (>= 9), autotools-dev, protobuf-compiler, libprotobuf-dev, dh-autoreconf, pkg-config, libutempter-dev, zlib1g-dev, libncurses5-dev, libssl-dev, bash-completion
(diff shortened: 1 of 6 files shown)


andersk:
This still won’t quite work: it seems we can’t guarantee the availability of `en_US.UTF-8` in a Debian build chroot without manually generating a copy of it in `debian/rules`! However, it is now the case that `C.UTF-8` comes with the base system in all supported Debian/Ubuntu releases (though it’s [not yet a thing upstream](https://sourceware.org/bugzilla/show_bug.cgi?id=17318)), so we probably just want to try both in some order.

cgull:
!!!
I was hoping to avoid `C.UTF-8` for the sake of simplicity and consistency, because it is not present on many other platforms.  More yak shaving tonight, or something.

cgull:
This might be an improvement:  it seems to work properly on Debian sid schroot, Ubuntu zesty, FreeBSD 11 and OS X.

cgull:
@andersk, @keithw, any comment?

keithw:
I don't know the "right" answer, but I can upload this to Debian experimental and see what happens...

cgull:
That got annoying.  It turned out that we are missing a build dep for `tmux` to get full testing, and adding that revealed more missing build deps and bugs.  I'll push this onto master once CI passes; @keithw, could you upload to Debian experimental at that point?

keithw:
Yes, happy to.

PR #862, opened 2017-03-20
Fix pledge() usage and make test cases work properly on OpenBSD

cgull:
Tested on OpenBSD 6.0.

This might or might not make it into mosh 1.3.0, but if not will be committed immediately after release.  I'm somewhat inclined to do this after release.  Jeremie, would it be possible for you to do this in the OpenBSD port for this release?  @keithw, do you have any opinion one way or another?

### DIFF
--- a/src/frontend/mosh-server.cc
+++ b/src/frontend/mosh-server.cc
@@ -577,7 +577,7 @@ static int run_server( const char *desired_ip, const char *desired_port,
     /* Drop unnecessary privileges */
 #ifdef HAVE_PLEDGE
     /* OpenBSD pledge() syscall */
-    if ( pledge( "stdio inet ioctl tty", NULL )) {
+    if ( pledge( "stdio inet tty", NULL )) {
       perror( "pledge() failed" );
       exit( 1 );
     }
--- a/src/frontend/stmclient.cc
+++ b/src/frontend/stmclient.cc
@@ -415,7 +415,7 @@ bool STMClient::main( void )
   /* Drop unnecessary privileges */
 #ifdef HAVE_PLEDGE
   /* OpenBSD pledge() syscall */
-  if ( pledge( "stdio inet ioctl tty", NULL )) {
+  if ( pledge( "stdio inet tty", NULL )) {
     perror( "pledge() failed" );
     exit( 1 );
   }
--- a/src/tests/e2e-test
+++ b/src/tests/e2e-test
@@ -60,6 +60,14 @@ test_exitstatus()
 # Tmux check.
 tmux_check()
 {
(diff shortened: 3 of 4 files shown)


keithw:
@cgull, what are the implications if we release 1.3.0 without this fix? Does Mosh still work on OpenBSD...? I'm a little unclear on what's broken in the current master.

cgull:
* A couple of weeks ago, Mosh built and ran on OpenBSD 6.0 and probably prior releases.  But as is, it would not run test cases that require tmux, because `tmux` on OpenBSD is not the portable version and does not have the -V flag to report a version number (I'd been working around it with a local build of the portable version of tmux).
* A week or so ago, I added code that tries to ensure a UTF-8 locale is available during tests.  That errors tests that use it on OpenBSD (and probably some other systems) because its `locale` command does not support the `charmap` argument (which is an extension beyond POSIX functionality).
* 3/18/2017, OpenBSD altered `pledge(2)` in a backwards incompatible way, dropping support for one of its possible arguments.  That change will be in OpenBSD 6.1 releasing in a couple of weeks.
* Jeremie worked up a patch to fix that problem.
* I tested this with `make check` on OpenBSD 6.0 and saw the locales problem, and also realized I could fix the tmux problem.  I only did this back to 6.0 because that's what I have installed and 5.9 goes out of support when 6.1 is released.

So, before these changes, mosh would build on any OpenBSD (but `make check` would fail), but would fail at startup on OpenBSD-current and the coming OpenBSD 6.1.  After these changes, `make`, `make check`, and Mosh itself will run correctly and completely on OpenBSD 6.0-6.1. They should also run correctly on earlier versions of OpenBSD, but without running the tmux-based tests.

cgull:
For testing the locale, I've realized that the better thing to do is write a wrapper program for our `is_utf8_locale()` function, which gets us the same test that Mosh does on all platforms.  That's probably not for 1.3.0, though.

keithw:
@cgull Honestly I'm inclined just to go with your recommendation on this one, but if you want my two cents, it seems like we may as well include this in 1.3.0 and credit our thorough release candidate process for catching this. :-) I'd rather that Mosh 1.3.0 work out-of-the-box on OpenBSD since they were kind enough to test our RC and report a bug.

cgull:
Hmm.  The `pledge()` change poses no risk to platforms other than OpenBSD, it's under `#ifdef PLEDGE`.  The other changes do pose regression risk for `make check` on other platforms.  So I think I will pull Jérémie's fix for that, and work on the other changes for post-release.

cgull:
This is all pulled now.

PR #866, opened 2017-03-26
Restore perl 5.8.8 support

achernya:
This is as variant of the patch I had in EL5 for a while, but forgot to upstream.
It turns out that the changes we made to support 5.10.1 are actually sufficient
for 5.8.8 if we remove unused imports, so remove them.


cgull:
Thanks!

PR #867, opened 2017-03-27
Work around JuiceSSH rendering bug

### LINKED ISSUE #846: juicessh not compatible with current mosh-server
JuiceSSH shows screen corruption when connecting to a mosh-server from master on OS X, running tmux 2.3 and WeeChat 1.5 (the original reporter saw this on Linux, I believe).  A 1.2.4 server works correctly in the same environment.  The original reporter's screenshot is here:  
http://imgur.com/JxV0Gn3

In my testing, it appears that when the screen is scrolled, graphic characters from previous lines are not erased, and the text of the newly-scrolled line overwrites the old text.  This appears to be a bad interaction between the new screen-update code in mosh-1.2.6 and JuiceSSH's proprietary code, but since JuiceSSH is closed source I can't address this further, and I've reported this to support@sonelli.com, and this is a tracking issue.

cgull:
JuiceSSH apparently has a bug where ECH for one character (ESC [ 1 X) does
not actually erase the character, in its code that receives and interprets
Mosh state updates.  This was hidden before because Mosh <= 1.2.5 never sent
this sequence, it sent ESC [ X instead as an optimization.

Do the better optimization of sending spaces for short sequences of blanks
instead.

This appears to fix #846.

### DIFF
--- a/src/terminal/terminaldisplay.cc
+++ b/src/terminal/terminaldisplay.cc
@@ -386,7 +386,7 @@ bool Display::put_row( bool initialized, FrameState &frame, const Framebuffer &f
       frame.append_silent_move( frame_y, frame_x - clear_count );
       frame.update_rendition( blank_renditions );
       bool can_use_erase = has_bce || ( frame.current_rendition == initial_rendition() );
-      if ( can_use_erase && has_ech ) {
+      if ( can_use_erase && has_ech && clear_count > 4 ) {
 	snprintf( tmp, 64, "\033[%dX", clear_count );
 	frame.append( tmp );
       } else {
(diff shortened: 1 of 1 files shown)


keithw:
It's been a while since I had this swapped in, but my recollection is that we always want to use ECH on systems that support it, even for short runs of blank cells, because there's a semantic difference (for copy-and-paste purposes) between overwriting a cell with a space character and erasing it. You may have more recent state though...

cgull:
This only affects what we do for blanks between two printing characters, not at the end of the line. Is it important there too?

keithw:
I don't know, to be honest, although I think you're right that it's fine within a line.

cgull:
Ah, comment confusion here. I think @keithw and I agree that ECH is unimportant mid-line, though.

lahwran:
looking forward to this fix! this has irritated me for a while. anything security-related preventing this from being merged into the nightlies?

cgull:
Pulled.  No, there wasn't anything security related stopping me, merely inattention and lack of focus on Mosh-- I've not been my best lately.

Issue #788, opened 2016-08-12
extra newlines on startup and exit


cgull:
2016-07-31 11:47:19     mva     by the way, mosh inserting some empty lines: in the shell, it is opening, the first line is empty, and all the output (including shell prompt) starting from second line. And after detaching there is also few empty lines between local command prompt with "mosh" command and "mosh is exiting" line
2016-07-31 11:48:29     mva     it adds some aesthetic discomfort :)
2016-08-01 01:11:17     cgull   mva:  I didn't notice your mention of odd start/stop behavior before, sorry.  Mosh uses alternate-screen mode by default, and it should start with a blank screen with a shell prompt on the first line.  It switches back to the normal screen mode on exit, and that should restore the screen as it was before Mosh, with a prompt and cursor on the next line.  Trying it now, I see what I expect on startup, but there
2016-08-01 01:11:17     cgull   are two extra blank lines on exit.  Can you get screenshots of what you're seeing and post them somewhere or open a Github issue?
2016-08-01 01:35:59     mva     cgull: how about screencast?
2016-08-01 01:36:31     cgull   that'd do nicely too.
2016-08-01 01:37:22     cgull   though I'm behind some very slow connectivity right now, might take a little while to download.  And I really should be going to sleep :)
2016-08-01 01:38:39     mva     there is 1.7M
2016-08-01 01:38:51     mva     not that much bigger than few screenshots ;)
2016-08-01 01:40:07     mva     https://mva.name/files/myscreencast.mp4
2016-08-01 01:51:57     mva     didn't you already went to sleep? :)
2016-08-01 02:00:15     cgull   And yet...and yet, I'm still awake.  I see your extra line at startup.  I don't see that locally.  I see your extra two lines at exit.  I do see that locally.  I'll look at it more tomorrow.  I'm really going to sleep now.  :)
2016-08-01 02:03:44     mva     good night :)

cgull:
I've now come up with two solutions to this problem, one kludgy and low-impact (`cleanup-newlines` branch, edit some blank lines out on output), and another that I think is a more correct solution but has some potential for problems (`no-pty` branch, don't use a pty for ssh, so we don't have echo on the ssh server and don't need to output a newline to deal with that).

I now feel more strongly that Mosh shouldn't be using a pty for ssh because of this issue.  But since shells will behave differently based on whether they have a tty or not, this will certainly change (and possibly break) Mosh's behavior for some people.

[...]

PR #868, opened 2017-03-30
Reduce excess newlines inside and outside Mosh session

cgull:
Mosh is adding an extra newline at the start of its full-screen session, and two newlines outside the full-screen session, which show up after exiting mosh.

We can eliminate the former newline (at the cost of some extra complexity), but the latter two newlines present a Hobson's choice:
* If we start ssh without a pty, we can eliminate all the newlines.  But that breaks backward compatibility with mosh <= 1.2.3.
* If we don't print any newlines before `MOSH CONNECT`, we run a risk of Mosh breaking for people who type ahead before mosh-client is started.
* We could run mosh-server with the programmatic equivalent of `stty -echo`.  But that also breaks backward compatibilty because older `mosh-server`s get the termios state from the ssh session.
* We can eat newlines in the `mosh` script, which eats newlines in any non-Mosh output it receives.

The compromise I chose here is to eat newlines in `mosh`.  Please contribute your opinions.

This is one of the things that argues for an incompatible break to `mosh2`, in my mind-- here, we'd use ssh without a pty, and with stdin redirected away from the client tty, so it doesn't eat typeahead.

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -423,6 +423,8 @@ if ( $pid == 0 ) { # child
       } else {
 	die "Bad MOSH CONNECT string: $_\n";
       }
+    } elsif ( m{^\s*$} ) {
+      # do nothing
     } else {
       if ( defined $port_request and $port_request =~ m{:} and m{Bad UDP port} ) {
 	$bad_udp_port_warning = 1;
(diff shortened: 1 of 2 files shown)


cgull:
Fixes #788.

PR #874, opened 2017-04-08
Make SysV STREAMS usage (for forkpty) more portable.

cgull:
Should resolve Haiku build issue.


cgull:
#875 fixed the Haiku build in a better way.

PR #873, opened 2017-04-07
Fix build on Haiku.

diversys:

Issue #872, opened 2017-04-06
mosh shows motd twice on illumos


pyhalov:
After fix for #190 was integrated mosh prints /etc/motd on illumos twice, as this is done by default /etc/profile. We use the following patch to fix it - https://github.com/pyhalov/oi-userland/blob/d6b8c501e991b96ae28b9d98672bcf92fe328ff3/components/shell/mosh/patches/01-do-not-print-motd.patch

cgull:
@pyhalov, could you test #875?

pyhalov:
Tested. Builds, runs, don't show motd twice.

timofonic:
@cgull @pyhalov Why is this still open then?

pyhalov:
Don't know. Closing.

PR #875, opened 2017-04-09
Various ports

cgull:
This should fix recent portability issues with OpenBSD, IlumOS/OpenIndiana, and Haiku.  This probably fixes issues in #862, #872, #873, #874.  Could I ask those of you involved with these ports to to test this?


cgull:
Rebased and pulled.

PR #877, opened 2017-04-12
Remove unused Action::operator==

andersk:


cgull:
I was stalling on pulling this because I wanted to rework Actions and clean up this whole area, but I still haven't figured out how to do that.  I give up for now, pulled.

Issue #881, opened 2017-05-03
Tables in man pages render with corruption


cgull:
On `#mosh` last night and today:
```irc
22:15 --> Hello71 (Hello71@wikipedia/Hello71) has joined #mosh
22:19 <Hello71> hm, "man nft" displays incorrectly in tmux via mosh for me
22:20 <Hello71> 1. mosh server tmux, 2. man nft, 3. press page down, expected results: man pages is displayed, actual results: tables are corrupted
22:20 <Hello71> works fine when paging up though
22:28 <achin> what version of mosh and tmux?
22:46 <Hello71> 1.3.0, 2.4
22:46 <Hello71> seems still broken if I resize window and restart man, but works fine with regular ssh or outside tmux
22:47 <achin> 1.3.0 on both the client and server?
22:48 <achin> i'm running 1.3.0 and tmux 2.2.  the tables in the nft man page seems OK to me
22:49 <Hello71> you're using UTF-8 line drawing?
22:50 <achin> yeah
22:51 <achin> sorry, i gotta head to bed.  hang around for others or open an issue on github
22:51 <Hello71> I think you need to have at least two tables in view for it to break
22:51 <Hello71> do you have a really small screen?
22:51 <achin> let me send you a screenshot
22:53 <achin> http://i.imgur.com/LRWdqFT.png
22:53 <achin> that was running "man nft" and then pressing spacebar twice
22:57 <Hello71> https://i.imgur.com/d9sZnY8.png
22:57 <Hello71> thanks anyways
23:00 <achin> but please do hang around or file a bug report.  this is worth diagnosing
23:03 <Hello71> try turning off your man colors or pressing pgup/pgdown randomly a few time
23:04 <Hello71> or just scroll down more
15:32 --> Hello71 (Hello71@wikipedia/Hello71) has joined #mosh
15:33 <Hello71> achin: back
16:25 --> Hello71 (Hello71@wikipedia/Hello71) has joined #mosh
16:41 <cgull> Hello71: took a look at this.  It works fine for me in Konsole, which is my best guess at the terminal program you're using...Oh but wait tmux.
16:43 <Hello71> cgull: same output in urxvt
16:48 <cgull> Now I'm getting it in Konsole -> mosh-devel -> Mac with tmux 2.4 -> ssh linuxmachine.
16:51 <Hello71> strange setup, but maybe frames need to be dropped
16:52 <Hello71> I have 100ms latency, didn't test localhost.
[...]

PR #883, opened 2017-05-06
Do not move cursor for SCROLL UP and SCROLL DOWN

cgull:
tmux 2.4 uncovered a bug in Mosh where it was leaving the cursor incorrectly positioned after the SCROLL UP and SCROLL DOWN commands (it moves the cursor along with the scrolled region, when it shouldn't move the cursor at all).  When displaying a man page with tables (or presumably any UTF-8 text with line-drawing characters), tmux will often draw the boxes first, then go back and fill the boxes with text.  Sometimes it uses the SCROLL UP command to position the screen correctly for the next box to draw.  tmux 2.3 was generally doing the same thing, but it was always doing an absolute reposition of the cursor after using those commands-- tmux 2.4 no longer does so.

This bug appears to have existed since the very early days of Mosh, I'm a little surprised the bug hasn't shown up before.

### DIFF
--- a/src/terminal/terminalframebuffer.cc
+++ b/src/terminal/terminalframebuffer.cc
@@ -119,11 +119,8 @@ void Framebuffer::scroll( int N )
 {
   if ( N >= 0 ) {
     delete_line( ds.get_scrolling_region_top_row(), N );
-    ds.move_row( -N, true );
   } else {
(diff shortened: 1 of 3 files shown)


cgull:
See #881 for an attached file that tickles the bug.

PR #885, opened 2017-05-07
Don't do prediction on large pastes into mosh-client

### LINKED ISSUE #482: Mosh's memory footprint becomes extremely large when pasting large inputs
I have 252KB in the paste buffer.

```
$ pbpaste | wc
    8064   44352  258108
```

When I paste this into a fresh mosh session, I see mosh-client's resident set size balloon to 5GB.

Environment:
- OS X 10.8.5 (Darwin damogran.local 12.5.0 Darwin Kernel Version 12.5.0: Sun Sep 29 13:33:47 PDT 2013; root:xnu-2050.48.12~1/RELEASE_X86_64 x86_64)
- mosh 1.2.4 from Homebrew on the client
- mosh 1.2.4 from mosh-1.2.4-1.fc14.x86_64.rpm on Fedora on the server

cgull:
Fixes #482.  See that for some more discussion.

### DIFF
--- a/src/frontend/stmclient.cc
+++ b/src/frontend/stmclient.cc
@@ -314,10 +314,18 @@ bool STMClient::process_user_input( int fd )
   if ( !network->shutdown_in_progress() ) {
     overlays.get_prediction_engine().set_local_frame_sent( network->get_sent_state_last() );
 
+    /* Don't predict for bulk data. */
+    bool paste = bytes_read > 100;
+    if ( paste ) {
+      overlays.get_prediction_engine().reset();
+    }
+
     for ( int i = 0; i < bytes_read; i++ ) {
       char the_byte = buf[ i ];
 
-      overlays.get_prediction_engine().new_user_byte( the_byte, local_framebuffer );
+      if ( !paste ) {
+	overlays.get_prediction_engine().new_user_byte( the_byte, local_framebuffer );
+      }
 
       if ( quit_sequence_started ) {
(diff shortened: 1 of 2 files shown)


cgull:
@keithw voiced some concern about disabling predictions on paste.  So I spent a little while playing with this patch on an actual bad link to see what its actual impact on prediction might be.  Reminder, this patch temporarily disables prediction and predictive echo for any paste >= 100 bytes.

* Any paste starting at the beginning of a line already has predictive echo disabled, so the patch has no effect there.
* Any paste that has a newline in it already disables predictive echo after the first line, so while the first line might flicker a bit with prediction, predictive echo will still be disabled at the *end* of the paste, so the patch has no significant effect there, either.
* I think the only case where this patch might negatively affect prediction is when you have typed the beginning of a line, enough echo has been received that predictive echo has been enabled, and you then paste 100+ bytes on a single line, with no newlines in the paste and no wrapping to the next line in the terminal.

I for one do not often paste in-line objects of that size, the best example I can think of is pasting URLs into text that I'm writing, and those are often set on new lines.  But there's many many ways to be using Mosh, does anyone else have a use case where this is important?  (Maybe for writing/editing copy-- but trying to do this on a laggy link is going to break a writer's flow in any case.)

cgull:
The above isn't quite right.  Any paste that does not end in a newline will have prediction disabled for any further manual typing on the same line.  So the effect is somewhat more significant than I thought.

cgull:
Pulled.  Comments on the behavior of this code, from anybody pasting into Mosh, will be *highly* appreciated.

PR #887, opened 2017-05-09
Version 1.3.1 rc1.  Update changelogs.

cgull:
@keithw, would you have a look at this and push it to master if you like what you see?


keithw:
Pushed and uploaded to Debian experimental.

Issue #657, opened 2015-08-02
On systems without IPv4 address mapping, mosh-server doesn't receive IPv4 by default


beam:
Hello, after upgrade to 1.2.5 i have problem run mosh server without -s param, 
there is **setsockopt( IP_RECVTOS ): Invalid argument** error.

```
root@freebsd:~ # uname -a
FreeBSD freebsd 10.1-RELEASE-p16 FreeBSD 10.1-RELEASE-p16 #0: Tue Jul 28 12:04:19 UTC 2015     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
r
root@freebsd:~ # mosh-server new
setsockopt( IP_RECVTOS ): Invalid argument

MOSH CONNECT 60001 PIuj3A9fIsVi326kdX8RJQ

mosh-server (mosh 1.2.5) [build mosh 1.2.5]
Copyright 2012 Keith Winstein <mosh-devel@mit.edu>
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

[mosh-server detached, pid = 98607]

Warning: termios IUTF8 flag not defined.
Character-erase of multibyte character sequence
probably does not work properly on this platform.
root@freebsd:~ # sockstat -4 | grep mosh
root@freebsd:~ # 
root@freebsd:~ # mosh-server new -s

MOSH CONNECT 60001 ZVmC9FR4bSb16I7XAW2aKw

mosh-server (mosh 1.2.5) [build mosh 1.2.5]
Copyright 2012 Keith Winstein <mosh-devel@mit.edu>
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

[mosh-server detached, pid = 11680]

Warning: termios IUTF8 flag not defined.
Character-erase of multibyte character sequence
probably does not work properly on this platform.
root@freebsd:~ # sockstat -4 | grep mosh
root     mosh-serve 11680 4  udp4   192.168.0.162:60001   *:*
root@freebsd:~ # 
```

after downgrade to 1.2.4 it works again

```
root@freebsd:~ # mosh-server new

MOSH CONNECT 60001 Me/iTk6y4osUhZQp57M5TQ

[...]

PR #891, opened 2017-05-22
Accept both IPv4 and IPv6 connections when listening on any address

muggenhor:
On some Linux systems this is the default. On most other operating
systems accepting IPv4 connections on a IPv6 socket only happens when
explicitly requested. Even on Linux it depends on the sysctl config.

This change makes it independent from the system's defaults whether IPv4
connections will be accepted too through the same socket.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -453,6 +453,11 @@ AC_CHECK_DECL([IP_RECVTOS],
      [Define if IP_RECVTOS is a valid sockopt.])],
   , [[#include <netinet/in.h>]])
 
+AC_CHECK_DECL([IPV6_V6ONLY],
+  [AC_DEFINE([HAVE_IPV6_V6ONLY], [1],
+     [Define if IPV6_V6ONLY is a valid IPv6 socket option.])],
+  , [[#include <netinet/in.h>]])
+
 AC_CHECK_DECL([__STDC_ISO_10646__],
   [],
   [AC_MSG_WARN([C library doesn't advertise wchar_t is Unicode (OS X works anyway with workaround).])],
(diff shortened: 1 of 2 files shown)


cgull:
Thanks!  This is actually the same issue as #657, I believe.  FreeBSD also has this socket option available so we can use this there too, and since the CI build succeeded, OS X must have it as well.  I believe we'll need to put it under an autoconf test for other systems though.

I'm wondering how none of us (but especially me) didn't notice this socket option before...

muggenhor:
I actually first encountered problems needing this change on FreeBSD. Then later on also on a Debian system where a sysctl was configured to make the default for this option to be turned on.

It should also compile and work from Windows XP upwards (I've used this option before, a long time ago on a software project that had to ran on XP). I've got no access to OSX myself though, but expect it to work the same as on the rest of the systems.

As for an autoconf test: I don't think that's necessary because these options are always macros, so it is just possible to wrap in an `#ifdef IPV6_V6ONLY` block.

muggenhor:
I've rebased on master and my last change adds a configure check.

cgull:
Pulled for 1.3.2, thanks.

Since OpenBSD doesn't support this socket option (they regard it as a security flaw and their kernel returns an error), `mosh-server` will abort when it is invoked without some option to specify/get the IP address to bind to.  However, `mosh` always uses this option unless you specify '--bind-server=any`, so in practice not a big problem there.  Still, should fix...

muggenhor:
But does it still provide the `IPV6_V6ONLY` macro at build time?

cgull:
Yes, it does, probably because of the POSIX requirement.

I was wrong in saying mosh-server will abort-- it actually does `perror()`, so an OpenBSD user may or may not be warned of this problem, but in any case mosh-server will continue.

PR #899, opened 2017-07-15
support osc 52 clipboard copy integration.

### LINKED ISSUE #637: mosh does not transmit OSC 52 sequence to client
Similar to issue #552, this limitation prevents me from copying text from remote shell sessions into my local terminal via [the OSC 52 escape sequence](https://github.com/macton/hterm#does-hterm-support-the-osc-52-aka-clipboard-operations-sequence).  It's also the only reason I still use SSH instead of Mosh. :scream:

I tried to patch this but got stuck. :cry:  Any hints on how to write a string to the raw terminal device? :pencil2: 

```diff
diff --git a/src/terminal/terminalfunctions.cc b/src/terminal/terminalfunctions.cc
index 35c8055..3c1b5e2 100644
--- a/src/terminal/terminalfunctions.cc
+++ b/src/terminal/terminalfunctions.cc
@@ -558,6 +558,25 @@ void Dispatcher::OSC_dispatch( const Parser::OSC_End *act, Framebuffer *fb )
        * 2: set window title */
       cmd_num = OSC_string[ 0 ] - L'0';
       offset = 2;
+    } else if ( (OSC_string.size() >= 3) && (OSC_string[ 2 ] == L';') ) {
+      /* OSC of the form "\033]XX;<Pt>\007" where XX can be:
+       * 10: Change VT100 text foreground color to Pt.
+       * 11: Change VT100 text background color to Pt.
+       * 12: Change text cursor color to Pt.
+       * 13: Change mouse foreground color to Pt.
+       * 14: Change mouse background color to Pt.
+       * 15: Change Tektronix foreground color to Pt.
+       * 16: Change Tektronix background color to Pt.
+       * 17: Change highlight background color to Pt.
+       * 18: Change Tektronix cursor color to Pt.
+       * 19: Change highlight foreground color to Pt.
+       * 46: Change Log File to Pt.  (This is normally disabled by a compile-time option).
+       * 50: Set Font to Pt.
+       * 51: reserved for Emacs shell.
+       * 52: Manipulate Selection Data.
+       */
+      cmd_num = 10 * (OSC_string[ 0 ] - L'0') + (OSC_string[ 1 ] - L'0');
+      offset = 3;
     }
     bool set_icon = (cmd_num == 0 || cmd_num == 1);
     bool set_title = (cmd_num == 0 || cmd_num == 2);
@@ -569,6 +588,9 @@ void Dispatcher::OSC_dispatch( const Parser::OSC_End *act, Framebuffer *fb )
 
       act->handled = true;
     }
+    else {
+      // HELP ME: how can I write the OSC_string as-is to the terminal device?
+    }
   }
 }
 ```

acornejo:
This feature is particularly relevant for mosh for chrome (which is also used in windows I believe). Closes #637 

PS: Someone should seriously consider running this codebase through clang-format, the coding style is all over the place, I did my best to be consistent with the existing style, but even simple things like tab/spaces are mixed in the codebase. If this is something the mantainers want, I would be happy to submit a pullrequest that passes the entire codebase through clang-format, together with a travis test that will fail for pull requests that don't follow the style.

### DIFF
--- a/src/terminal/terminaldispatcher.cc
+++ b/src/terminal/terminaldispatcher.cc
@@ -42,6 +42,8 @@
 
 using namespace Terminal;
 
+static const size_t MAXIMUM_CLIPBOARD_SIZE = 16*1024;
+
 Dispatcher::Dispatcher()
   : params(), parsed_params(), parsed( false ), dispatch_chars(),
     OSC_string(), terminal_to_host()
@@ -235,7 +237,7 @@ void Dispatcher::dispatch( Function_Type type, const Parser::Action *act, Frameb
 void Dispatcher::OSC_put( const Parser::OSC_Put *act )
 {
   assert( act->char_present );
-  if ( OSC_string.size() < 256 ) { /* should be a long enough window title */
+  if ( OSC_string.size() < MAXIMUM_CLIPBOARD_SIZE) {
     OSC_string.push_back( act->ch );
   }
 }
--- a/src/terminal/terminaldisplay.cc
+++ b/src/terminal/terminaldisplay.cc
@@ -68,13 +68,13 @@ std::string Display::new_frame( bool initialized, const Framebuffer &last, const
   if ( f.get_bell_count() != frame.last_frame.get_bell_count() ) {
     frame.append( '\007' );
   }
+  typedef Terminal::Framebuffer::title_type title_type;
 
   /* has icon name or window title changed? */
   if ( has_title && f.is_title_initialized() &&
        ( (!initialized)
          || (f.get_icon_name() != frame.last_frame.get_icon_name())
          || (f.get_window_title() != frame.last_frame.get_window_title()) ) ) {
-    typedef Terminal::Framebuffer::title_type title_type;
       /* set icon name and window title */
     if ( f.get_icon_name() == f.get_window_title() ) {
       /* write combined Icon Name and Window Title */
@@ -108,6 +108,18 @@ std::string Display::new_frame( bool initialized, const Framebuffer &last, const
       frame.append( '\007' );
     }
 
+  }
+
+  /* has clipboard changed? */
+  if (f.get_clipboard() != frame.last_frame.get_clipboard()) {
+    frame.append( "\033]52;c;" );
+    const title_type &clipboard( f.get_clipboard() );
+    for ( title_type::const_iterator i = clipboard.begin();
+          i != clipboard.end();
+          i++ ) {
+      frame.append( *i );
+    }
+    frame.append( '\007' );
   }
 
   /* has reverse video state changed? */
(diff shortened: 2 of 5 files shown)


cgull:
Sorry to not respond to this earlier, I got reminded of this when looking at mosh-chrome issues.

This actually looks reasonable, and I'm considering pulling it for the next version of Mosh.  The one thing I want to think about is that a 100KB paste is a relatively large object to be passing in a single SSP message, and I see a bit of potential for excessively-large messages getting packets dropped by too-short network queues, causing the session to stall.

Mosh has only ever promised to deliver a faithful, usable terminal window to the client; it's never promised to deliver any particular sequence of bytes reliably from server to client.  This patch would bend that slightly, and it does require some arbitrary size limit (100KB actually seems fairly reasonable).

keithw:
Could we make it 16 KiB? :-) I'd feel more comfortable with something in
that range. A single 100 KiB Instruction is going to perform really poorly
on a network with any significant loss, since Mosh basically requires every
fragment to get there to make forward progress.

acornejo:
Thanks for the feedback. Changed max clipboard size to 16k, I think that is more than enough for most uses of the clipboard.

alexbozhenko:
Hi, will this patch resolve the following:
MacOs Sierra 10.12.5 ->
ITerm2 3.0.15 -> "Applications in terminal may access clipboard" checkbox is set in preferences-general ->
**ssh** to centos 7 server -> start tmux 2.2 -> copy some text to tmux buffer -> this text also copied to mac clipboard 

If I change ssh to **mosh**, tmux buffer doesn't appear in mac clipboard. Looks like tmux is using osc 52 to do that? If this patch would fix that, that would be awesome)

acornejo:
Yup, it will fix that too.

alexbozhenko:
Wow, cool. Sorry, dumb question: do I need to patch both server and client side to make it work? I want to try to build it

alexbozhenko:
Ok, I tried to build just client, and connecting with new client to old mosh server did not fix the issue with tmux clipboard that I described before.

acornejo:
If the new server is not sending you clipboard data I will be surprised if
the client works.

cgull:
Pulled.  Thanks!

cgull:
@acornejo I forgot to follow up about the source-code formatting:

Where do you find the source code style to be badly inconsistent?  I find it mostly consistent with Emacs formatting set for cc-mode, with 8-character tab stops, and KeithW's odd style of including spaces in parentheses in most contexts.  ocb.cc is a major exception, because that was copied into the project, and there's value in retaining that security-critical code as contributed, for ease of comparison with the original and other forks.  There's a few other files that are a little inconsistent but I haven't found it a major issue.

I personally would like to reformat the source (and get rid of the spaces around the parens, which really annoy me), but that's a project-wide decision.

I'd certainly like to see your clang-format machinery, or at least the clang-format invocation you use, to see what we might get-- I haven't used it much myself, and not for quite a while.

dohsimpson:
> Ok, I tried to build just client, and connecting with new client to old mosh server did not fix the issue with tmux clipboard that I described before.

Yep, you need to build the server and the client.

PR #900, opened 2017-07-17
mosh 1.3.1 proposed

cgull:
Run CI on what will? should? be mosh-1.3.1.  Also let others (@keithw in particular) comment on the wisdom of including some code changes in the final release.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,8 @@
 	* Version 1.3.1 released.
 
 	* Platform support:
+		* Explicitly enable binding to both IPv4 and IPv6 addresses.
+		  (Giel van Schijndel)
 		* Restore perl 5.8.8 support for RHEL5.  (Alexander Chernyakhovsky)
 		* Make tests detect UTF-8 locale with a helper executable.  (John Hood)
 		* Don't print /etc/motd on IllumOS.  (John Hood)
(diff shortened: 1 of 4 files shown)


keithw:
lgtm

nckx:
Hi! Is 1.3.1 now considered a final (and official) release? I'm asking because of this open PR, and because mosh.org still only hosts 1.3.0 after a few days.

cgull:
Yes, it'll be 1.3.2, thanks to my error in debian/changelog.  I'm deleting the 1.3.1 tag and Github release now.

nckx:
Thanks! I'll just wait for 1.3.2 to appear on mosh.org then; GitHub's tarballs for previous releases don't match the official sha256sums.

cgull:
It isn't a release till we send the email and update https://mosh.org.  (Perhaps we should better automate/atomicize this, but that's what we do now.)

The checksums on dist tarballs on GitHub for previous *releases* (1.2.5, 1.2.6, 1.3.0) match for me, so I'm not sure what you're seeing.  Perhaps you're looking at release candidates, which we don't sign and may not be as careful with?   In any case, our authoritative distribution for releases is from mosh.org, not GitHub.

nckx:
I'm glad to hear that. It's not always that clear.

*About those checksums:
I downloaded 1.2.x – definitely a release – from GitHub a while back, extracted & built it, then noticed that the hash was off. Maybe the tarball was truncated, and I missed the error? I didn't think much of it. Many projects do weird or at least different things on GitHub than in the real world.
I don't have that tarball now, and `curl | sha256sum` returns identical sums like expected, so I've no idea.* \*shrugs\*

glensc:
i guess he downloaded the github generated tarball, because the uploaded tarball was not present. @nckx include url what you downloaded (next time?) in your report! :)

nckx:
Your guess is as good as mine. If I knew, I would. Anyway, on to newer and better tarballs!

keithw:
The release has now happened. The source code is at: https://mosh.org/mosh-1.3.2.tar.gz
(SHA-256: da600573dfa827d88ce114e0fed30210689381bbdcff543c931e4d6a2e851216)

Issue #188, opened 2012-04-16
Streamline server install from client connection


ghost:
Scenario:
User has the mosh client installed and wants to connect with it to a server that doesn't have mosh-server installed. Furthermore, user doesn't have admin privileges on server.

Desired:
running 'mosh server-location" looks for the mosh-server binary. If none found, it prompts the user to install mosh on the server locally. Then the mosh-server binary for the current architecture is downloaded and stored locally (say "~/.mosh/mosh-server" or "~/.local/share/mosh/mosh-server") and mosh-server is started.

Current:
To install a workaround is needed which is a huge pain.

Workaround (all plaforms):
Connect through ssh.
Install mosh locally on the server:
  Get all dependencies (ain't happening.. not without admin):
    Protocol Buffers    protobuf-compiler, libprotobuf-dev
    Boost   libboost-dev
    utempter    libutempter-dev
    ncurses libncurses5-dev
    zlib    zlib1g-dev
    IO::Pty Perl module libio-pty-perl
  $ wget https://github.com/downloads/keithw/mosh/mosh-1.1.3.tar.gz
  $ tar -zxvf mosh-1.1.3.tar.gz
  $ cd mosh-1.1.3
  $ ./configure
  $ make
  (mosh-server is in .src/frontend/make-server)
exit
mosh --sever=/path/to/mosh-server server-location

Workaround Ubuntu 11.10 (or for other distributions go to https://launchpad.net/~keithw/+archive/mosh/+packages and wget the correct .deb distribution package):
Connect through ssh.
  $ wget https://launchpad.net/~keithw/+archive/mosh/+files/mosh_1.1.3-0~585~oneiric1_amd64.deb
  $ ar -vx /mosh_1.1.3-0~585~oneiric1_amd64.deb
  $ tar -zxvf data.tar.gz
  $ cp /usr/bin/mosh-server .
  $ exit
mosh --sever=/path/to/mosh-server server-location

notorand-it:
I think this is not really doable. And shouldn't be implemented.

**First**, I think that installing any software automatically at the connection is going to be considered as an attempt to circumvent the system administration policies. Unless explicitly allowed. The proper way is you to ask the sysadmins to install mosh.

[...]

PR #906, opened 2017-07-26
Adds Dockerfile-s for building a static mosh-server on Alpine Linux

javabrett:
Prototype dockerfiles to build/test a statically-linked (musl glibc) mosh-server.  Part of research on #188.
- Dockerfile - Alpine linux base, installed build toolchain, checks-out mosh master, builds/installs, starts sshd and creates a mosh test-user.
- Dockerfile.ubuntu - Ubuntu base running sshd with mosh test-user.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -1,0 +1,22 @@
+FROM alpine:latest
+MAINTAINER Brett Randall <javabrett@gmail.com>
+
+# NOTE: openssh is only required for testing connection to built docker using the built mosh-server
+RUN apk update \
+	&& apk add alpine-sdk autoconf automake protobuf-dev ncurses ncurses-dev ncurses-static openssl-dev openssh
+
+RUN git clone https://github.com/mobile-shell/mosh.git \
+	&& cd mosh \
+	&& ./autogen.sh \
+	&& LDFLAGS="-static" ./configure --prefix=/usr \
+	&& make \
+	&& make install \
+	&& ssh-keygen -f /etc/ssh/ssh_host_rsa_key -N '' -t rsa \
+	&& ssh-keygen -f /etc/ssh/ssh_host_dsa_key -N '' -t dsa \
+	&& adduser -D mosh \
+	&& echo 'mosh:m0sh-st@t1c' | chpasswd
+
+EXPOSE 22
+
+ENTRYPOINT /usr/sbin/sshd \
+	&& sh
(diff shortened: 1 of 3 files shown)


eminence:
I just tested this out.  The build seemed to be fine, but when I tried to copy the static `mosh-server` binary to my gentoo machine, it segfaults when I run it.  Does this happen for you?

javabrett:
@eminence I had not tested the static `mosh-server` with Gentoo, only Ubuntu.  So I just created a Gentoo test-target dockerfile and added that.  From my testing on this image, the same static `mosh-server` worked ok, so no segfault.

Can you repeat the test using `Dockerfile.gentoo` as the host for the static `mosh-server` and report whether it works for you?  Should do, and if-so would need more details on the target Gentoo host that you tested on, details of the segfault etc.

eminence:
I tested the musl-static `mosh-server` in the image created by `Dockerfile.gentoo` and it still segfaults.  

Here's what gdb has to say (running on my gentoo host, not a gentoo docker guest):

```
Program received signal SIGSEGV, Segmentation fault.
memcpy () at src/string/x86_64/memcpy.s:18
18      src/string/x86_64/memcpy.s: No such file or directory.
(gdb) bt
#0  memcpy () at src/string/x86_64/memcpy.s:18
#1  0x000055555565065c in __copy_tls (mem=0x5555558d0438 <builtin_tls+56> "") at src/env/__init_tls.c:60
#2  0x00005555556507a2 in static_init_tls (aux=aux@entry=0x7fffffffdc00) at src/env/__init_tls.c:127
#3  0x0000555555644877 in __init_libc (envp=envp@entry=0x7fffffffdda0, pn=<optimized out>) at src/env/__libc_start_main.c:38
#4  0x0000555555644991 in __libc_start_main (main=0x55555557c2a0 <main(int, char**)>, argc=2, argv=0x7fffffffdd88) at src/env/__libc_start_main.c:70
#5  0x000055555558017f in __dls2 (base=<optimized out>, sp=<optimized out>) at crt/rcrt1.c:14
#6  0x000055555558000d in _start ()
```

javabrett:
@eminence thanks for testing and the stack-trace.

Unfortunately I've not been able to reproduce this yet.  Similar crashes in bugtrackers (mainly for Alpine) for musl glibc seem to indicate that it could relate to a much smaller default stack size in musl.

Not sure what to suggest next.  Is it easy to compile a `mosh-server` static within a Gentoo image?  I'm not as familiar with Gentoo images or dev tool-chain.  Maybe a statically-linked version built on Gentoo will run there.  I did hope that the binary would be fairly portable though.

javabrett:
I'm going to close this PR - other discussions e.g. https://github.com/mobile-shell/moshweb/pull/47 lead me to believe that it is better to pursue dynamically-linked options with relocated/local dependencies as a non-root install-option, that the difficulty of a (portable?) statically-linked build.

Thanks for the testing anyway.

PR #907, opened 2017-08-01
Automatic memory

cgull:
These changes nearly eliminate manual memory management (`new`/`delete`, `malloc`/`free`, C strings) from Mosh.  Most of these changes are simplifications and will not be controversial.  There's a couple of places where `unique_ptr`/`scoped_ptr` would be more appropriate.  I used `shared_ptr` instead because the performance impact in those cases was small and `unique_ptr`/`scoped_ptr` are not reliably available in Mosh's C++03/TR1 environment.

The one somewhat-arguable thing is the rework to use `shared_ptr` to track `Action`s.  This change does slow down Mosh's terminal emulator input slightly, current testing suggests by about 20% on high-speed output to mosh-server.  That slowdown should disappear for typical usage, where small updates are much more common. `shared_ptr` is rather heavyweight for this, but since `Action`s are message objects to begin with and don't have a single point of ownership, nothing simpler works.


cgull:
Pulled.

PR #924, opened 2017-08-24
Support for libgcrypt, and external OCB in OpenSSL and libgcrypt

cgull:
This supersedes #790.

It adds support for libgcrypt's AES implementation.  It also supports the OCB implementations in libgcrypt and OpenSSL 1.1+, in place of the bundled Rogaway/Krovetz implementation.  There's autoconf support for all these permutations.  Unlike #790, the autoconf code defaults to the current behavior, and libgcrypt and/or external OCB must be manually selected.

The external-OCB support is done by adding shims that support the `ae_encrypt`/`ae_decrypt` API presented by the bundled code.  The shims are incomplete and do not support the full range of possible OCB encryption/decryption actions available in that API, or the full range of OCB cipher configurations.  However, they do support the straightforward OCB requests that Mosh makes, and likely many of the requests that other practical OCB-using programs will make.  The code passes all of Mosh's OCB tests, but does not pass the Krovetz/Rogaway test code at the bottom of `src/crypto/ocb.cc`.  I think this code might be useful for other projects even if Mosh never uses it.

The libgcrypt support has seen some testing on Linux, FreeBSD, and macOS.  The OpenSSL OCB code has only been tested on FreeBSD and needs verification on other platforms, once they take up OpenSSL 1.1+ more seriously.  So this is not yet really ready to pull into Mosh.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -346,6 +346,43 @@ case "${with_crypto_library}" in
     ;;
 esac
 
+AC_ARG_WITH(
+  [external-ocb],
+  [AS_HELP_STRING([--with-external-ocb], [Use the OCB implementation from the selected crypto library @<:@no@:>@])],
+  [with_external_ocb="$withval"],
+  [with_external_ocb="no"])
+
+  AS_IF([test x"$with_external_ocb" != xno],
+    [
+      # push LDFLAGS/CFLAGS
+      old_LIBS="$LIBS"
+      old_CFLAGS="$CFLAGS"
+      # modify
+      LIBS="$LIBS $CRYPTO_LIBS"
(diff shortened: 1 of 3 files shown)


cgull:
This work has been mostly superseded by newer, better OpenSSL OCB support in #1196 by @achernya, and later work, included (but not default) in Mosh 1.4.0.  The libgcrypt support here might still be of interest to some, but given libgcrypt's unpopularity, isn't worth pulling into Mosh.

After the problems discovered with OpenSSL's OCB, I now better understand @keithw's conservative position on using outside crypto implementations, and mostly agree with it.  I think alternate crypto implementations still have some value for development and testing.

PR #933, opened 2017-10-24
update

foresthz:


mm
```diff
@@ -0,0 +1,11 @@
+comment="m"
```
glensc: missing shebang definition. most portable way is using env:

```
#!/usr/bin/env sh
```

mm
```diff
@@ -0,0 +1,11 @@
...
+if [ -z "$1" ]; then
+echo "=== default comment";
```
glensc: unnecessary command terminator. drop `;`

glensc: if true branch should be indented. my style is to use tabs

mm
```diff
@@ -0,0 +1,11 @@
...
+else
+echo "=== new comment";
+echo $1;
```
glensc: surround `$1` with quotes to avoid unexpected shell glob

mm
```diff
@@ -0,0 +1,11 @@
...
+comment=$1;
```
glensc: surround `$1` with quotes to be portable, some shells may expand globs here

glensc:
overall comment: the change doesn't belong to this repository. good luck! :D

PR #934, opened 2017-10-24
Add syslog logging of connections

amishHammer:
Log connection change events to syslog in the auth log, logging the PID,
username and remote host.


src/frontend/mosh-server.cc
```diff
@@ -657,6 +664,12 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
   socklen_t saved_addr_len = 0;
   #endif
 
+  #ifdef HAVE_SYSLOG
+  Addr syslog_saved_addr;
+  socklen_t syslog_saved_addr_len = 0;
+  openlog("mosh-server", LOG_PID | LOG_NDELAY, LOG_AUTH);
```
cgull: `argv[0]`, maybe?

Also, please move the `openlog()` to `run_server()`, before the `pledge()` call.  That's the break between normal operation and sandboxed operation.  (It doesn't actually matter for OpenBSD, which now has a `sendsyslog()` syscall to handle this and other issues, but the principle remains, for future sandboxing efforts.)

src/frontend/mosh-server.cc
```diff
@@ -783,6 +796,28 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
...
+	    char host[ NI_MAXHOST ];
+	    int errcode = getnameinfo( &syslog_saved_addr.sa, syslog_saved_addr_len,
+				       host, sizeof( host ), NULL, 0,
+				       NI_NUMERICHOST );
+	    if ( errcode != 0 ) {
+	      throw NetworkException( std::string( "serve: getnameinfo: " ) + gai_strerror( errcode ), 0 );
+	    }
+        struct passwd *pw = getpwuid( getuid() );
+        if (pw == NULL) {
+	      throw NetworkException( std::string( "serve: getpwuid: " ) + gai_strerror( errno ), 0 );
```
cgull: You want `strerror()` or `strerror_r()` here.

src/frontend/mosh-server.cc
```diff
@@ -783,6 +796,28 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
 	    connected_utmp = true;
 	  }
 	  #endif
+      #ifdef HAVE_SYSLOG
+	  if ( syslog_saved_addr_len != network.get_remote_addr_len()
+	       || memcmp( &syslog_saved_addr, &network.get_remote_addr(),
```
cgull: My eyebrows went up at the use of `memcmp()` on what looked like a C++ object, but sadly I'm not sure we can actually do any better here, since `Addr` is actually a union of C structs.

The best thing would be to have `Addr` be a variant object with an `operator==`, but I'm not asking you to do that.

src/frontend/mosh-server.cc
```diff
@@ -657,6 +664,12 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
   socklen_t saved_addr_len = 0;
   #endif
 
+  #ifdef HAVE_SYSLOG
+  Addr syslog_saved_addr;
```
cgull: I think you can reuse the existing `saved_addr` and `saved_addr_len` variables, rather than duplicating them, even with the complication of `connected_utmp`.

src/frontend/mosh-server.cc
```diff
@@ -52,6 +52,13 @@
 #ifdef HAVE_UTEMPTER
 #include <utempter.h>
 #endif
+#ifdef HAVE_SYSLOG_H
+#ifdef HAVE_STDARG_H
```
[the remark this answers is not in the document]
amishHammer: Will fix this one.

src/frontend/mosh-server.cc
amishHammer: You can't because on address change in the connected_utmp case, the new addresses will be populated into saved_* before the utmp block ends meaning they will be set to the current connection address by the time we get to the syslog block.

src/frontend/mosh-server.cc
amishHammer: Ok will fix this.

src/frontend/mosh-server.cc
amishHammer: This is the same as the utmp_connected case, minus the condition on utmp_connected.  If one changes both cases should.

src/frontend/mosh-server.cc
amishHammer: Will fix this.

amishHammer:
These changes where requested by the security team at my employer.  Repeatedly dumping the contents of utmp is not a viable solution to the problem.  They want to see log entries in the auth log from the mosh-server that tell them what IP/host a user is connecting from and if/when this user changes IP they want a log entry for when that happens and what the new address is.  Repeatedly dumping utmp would lead to race conditions and other scenarios that are best solved by logging the connection information.  I would be happy hiding this behind a configure flag to enable it if there was concern about it being on by default.

feld:
sshd logs connections to syslog with the auth/authpriv facility so mosh should as well so the audit trail is of interactive login sessions is clearly accounted for

cgull:
@feld's comment points out two things for me:
* A Mosh session is already logged as its initiating ssh session.
* sshd logs as auth/authpriv because it's authenticating a session.  When users authenticate, there is the risk of passwords getting typed at login prompts, so authentication requests should be logged in a tightly secured file.  Mosh, of course, does not authenticate.  OTOH, auth/authpriv might be a logical grouping for this info (much other session info goes there).  So I'm not sure what the right syslog facility/level to use is.

src/frontend/mosh-server.cc
cgull: So, refactor the blocks :)

configure.ac
```diff
@@ -213,6 +213,7 @@ AC_CHECK_HEADERS([endian.h sys/endian.h])
 AC_CHECK_HEADERS([utmpx.h])
 AC_CHECK_HEADERS([termio.h])
 AC_CHECK_HEADERS([sys/uio.h])
+AC_CHECK_HEADERS([stdarg.h syslog.h])
```
cgull: Since syslog.h and the syslog family of calls are in POSIX, we don't need an autoconf header check.  (We might need a define check for auth/authpriv if we decide to go that route.)

Can you put this under a configure option, though?

amishHammer:
@cgull while the initial connection is already logged, it is the hopping connections that we (@feld and myself) are trying to get logged.  We need a record of the IP's the users session hops between not just the initial connection.  These latest commits should clear up most of your issues.

keithw:
If people want syslog logging (at least behind an optional compile-time or runtime flag), that's fine with me. I should point out that you don't need to repeatedly dump utmp or worry about race conditions if you don't dump it frequently enough -- every change of IP address is already logged to wtmp, and you can see these events in human-readable format with tools like `last`.

configure.ac
```diff
@@ -177,6 +177,17 @@ AS_IF([test x"$with_utempter" != xno],
       [AC_MSG_WARN([Unable to find libutempter; utmp entries will not be made.])],
       [AC_MSG_ERROR([--with-utempter was given but libutempter was not found.])])])])
 
+AC_ARG_ENABLE([syslog],
+  [AS_HELP_STRING([--enable-server], [Enable connection logging in mosh-server @<:@yes@:>@])],
```
cgull: I think you mean `--enable-syslog` here.  Also, I think I want this to default to no for now.

src/frontend/mosh-server.cc
```diff
@@ -769,14 +767,25 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
 	  if ( !network.shutdown_in_progress() ) {
 	    network.set_current_state( terminal );
 	  }
-	  
-	  #ifdef HAVE_UTEMPTER
-	  /* update utmp entry if we have become "connected" */
-	  if ( (!connected_utmp)
+      #if defined(HAVE_SYSLOG) || defined(HAVE_UPTEMPTER)
+      #ifdef HAVE_UTEMPTER
+      if (!connected_utmp) {
```
cgull: There are indentation problems here (and the next hunk).  Mosh source code is generally written in GNU style, using both spaces, and tabs at 8-character stops for indentation.  This code looks like it was edited with 4-character tab stops.

src/frontend/mosh-server.cc
```diff
@@ -789,34 +798,22 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
 	      throw NetworkException( std::string( "serve: getnameinfo: " ) + gai_strerror( errcode ), 0 );
 	    }
 
+        #ifdef HAVE_UTEMPTER
```
cgull: The refactoring has made it obvious that the syslog code doesn't log lost connectivity.  I'm guessing you don't need this for your auditing needs, but would you consider logging lost connectivity as well?

src/frontend/mosh-server.cc
```diff
@@ -789,34 +798,22 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
...
+        utempter_remove_record( host_fd );
 	    char tmp[ 64 ];
 	    snprintf( tmp, 64, "%s via mosh [%d]", host, getpid() );
 	    utempter_add_record( host_fd, tmp );
-
```
cgull: And this line is removed whitespace compared to master.

src/frontend/mosh-server.cc
```diff
@@ -659,16 +660,13 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
...
   socklen_t saved_addr_len = 0;
   #endif
 
-  #ifdef HAVE_SYSLOG
-  Addr syslog_saved_addr;
-  socklen_t syslog_saved_addr_len = 0;
-  openlog("mosh-server", LOG_PID | LOG_NDELAY, LOG_AUTH);
-  #endif
 
   bool child_released = false;
```
cgull: This line is extra whitespace compared to Mosh master.

src/frontend/mosh-server.cc
cgull: ...and we might want to log connection close as well, at the end of `run_server()`.

src/frontend/mosh-server.cc
```diff
@@ -278,6 +281,11 @@ int main( int argc, char *argv[] )
 
   bool with_motd = false;
 
+  #ifdef HAVE_SYSLOG
+  openlog(argv[0], LOG_PID | LOG_NDELAY, LOG_AUTH);
```
cgull: We should also do `closelog()` in the `forkpty()` child.

cgull:
I pushed something that covers the points I mentioned and a few more.

I am undecided whether this feature should be enabled at build time or at run time.  Enabling at build time gives more control to people building Mosh, but it also means we have significant functionality that is not always enabled/exercised.  On the other hand, there's some real value in keeping Mosh's interaction with the host system simple, from a security and portability view.

feld:
Is this going to make it into 1.3.3?

achernya:
Closing due to lack of activity and unresolved code review comments.

Issue #935, opened 2017-10-25
Compiled and installed sucessfully, but failed to execute mosh


c02y:
- CentOS6.5-32bit

I downloaded mosh-1.3.2.tar.gz file from page https://mosh.org/#getting and followed the instructions to compile and install mosh from the source code. The compilation and installation works smoothly, but when I tried to execute the mosh binary file like `mosh --help` or `mosh`, error messages occurred to me:

    Bareword "AI_NUMERICHOST" not allowed while "strict subs" in use at /usr/bin/mosh line 489.
    Bareword "AI_CANONNAME" not allowed while "strict subs" in use at /usr/bin/mosh line 498.
    Bareword "NI_NUMERICHOST" not allowed while "strict subs" in use at /usr/bin/mosh line 336.
    Bareword "NI_NUMERICHOST" not allowed while "strict subs" in use at /usr/bin/mosh line 261.
    Execution of /usr/bin/mosh aborted due to compilation errors.

eminence:
What version of perl do you have?

cgull:
According to 5ee542c we should now support all the way back to RHEL5, so this is surprising.  Are you using the distribution‘s Perl?

c02y:
@eminence 
> What version of perl do you have?

v5.10.1 (*) built for i386-linux-thread-multi

@cgull 
> Are you using the distribution‘s Perl?

Yes.

cgull:
When building from source, you'll need to supply all the dependencies before Mosh will build and work.  See `fedora/mosh.spec`.   On Centos 6.5, the `protobuf-compiler`, `protobuf-devel`, and `perl-IO-Socket-IP` packages aren't available from standard repos (but you've evidently gotten the protobuf dependencies from somewhere, or you wouldn't have gotten this far).

It seems that the `mosh` script is mistakenly thinking it's imported useful getaddrinfo support and then failing on the subsequent Perl compile step, that's a (minor) Mosh bug.

cgull:
@andersk: on CentOS 6.5, the `eval { Socket::GetAddrInfo->import( '0.22', @gai_reqs ); 1` succeeds but getaddrinfo isn't actually available-- got any insights?

andersk:
You are missing `Socket::GetAddrInfo`. The error message that was intended to clarify this was broken, and I’ve fixed it in #940.

cgull:
[...]

PR #937, opened 2017-10-26
Add Perl compile

cgull:
This helps catch Perl syntax/version/dependency issues at build time
rather than run time.  Addresses the issue brought up in #935.

### DIFF
--- a/scripts/Makefile.am
+++ b/scripts/Makefile.am
@@ -5,5 +5,6 @@ endif
 CLEANFILES = $(bin_SCRIPTS)
 
 mosh:	mosh.pl ../VERSION Makefile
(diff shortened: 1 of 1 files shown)


eminence:
I confirm this code works on a RHEL7.3 machine with perl v5.8.8

But note that the error came during the compilation step (i.e. `make`), not `./configure`.   Is there any way to bring the error into the `./configure`?  Maybe this is just me, but I generally expect dependency checks to be done in the configure step?   If not, can we produce an error that might suggest this problem is due to perl versions?

cgull:
Technically that might make some sense, but I'm content to call a working Socket::GetAddrInfo a required build/run dependency for Mosh and have broken platforms fail during compile, about 10 seconds later than they would have if we had a Configure test.

Issue #649, opened 2015-07-13
Add support for 24-bit color


agriffis:
Lots of terms and related programs are gaining support for 24-bit color escapes, see https://gist.github.com/XVilka/8346728

I'd love to see support for this in mosh. Creating this issue to gather information on what's needed in mosh, and to coordinate any effort around it.

keithw:
No fundamental objection to supporting this, but what terminfo entry are these in?

Right now the `mosh-client` gets the number of client colors with `tigetnum( "colors" )` (see https://github.com/keithw/mosh/blob/master/src/frontend/mosh-client.cc#L87), and then based on that answer, the `mosh-server` ends up advertising either TERM=xterm or TERM=xterm-256color to the host applications.

What TERM can `mosh-server` advertise to indicate that it supports these truecolor escape sequences, and how can `mosh-client` determine if the client supports them?

cgull:
This is all quite doable, but there are issues to work through:

This adds a *fourth* set of rendering attributes to character cells (VT100, 16-color, 256-color, true-color).  Since Mosh is a virtual terminal that uses an unknown real terminal for its final rendering, I think it has to maintain and pass through all four sets of attributes.  16-color attributes can't be converted to RGB because different terminals choose different colors, often user configurable.  256-color attributes could be converted to RGB, but I think performance and compatibility issues will weigh against that.

I'm a bit concerned about performance.  Adding support for this will add 6+ bytes to each Terminal::Framebuffer::Cell, but more importantly, the escape sequences are verbose (14 characters minimum to set just foreground or background, 21 to set both).  We will want to pass as few of these as possible.  256-color attributes are a third shorter than this and should be passed as such.

Implementation for this will touch the same code as my upcoming performance changes.

What do we do with a terminal emulator that doesn't support true-color mode?  Passing this stuff through to a truly ignorant terminal will result in *RGBake* (you get the right characters, tossed in a fruit salad of colors/attributes).  We can either ignore the problem, or we can do truecolor detection similar to Mosh's existing 16/256-color detection.

If we don't implement this soon, we should at least implement detection of these sequences so that we properly ignore them.

<https://groups.google.com/forum/#!msg/iterm2-issues/NXLGvwuzjrc/35gVs4pfjLYJ> has a fair bit of relevant commentary.

cgull:
I spent a while gnawing on this last night.  It's complicated, and I found we're not currently fully/correctly supporting 256-color attributes, either.

[...]

Issue #507, opened 2014-03-24
Escape codes that change color don't work properly


pankdm:
It seems like mosh doesn't support escape sequences that contain both foreground and background color changes (see http://en.wikipedia.org/wiki/ANSI_escape_code#Colors).
The following command isn't shown correctly in mosh:
```
echo -e "\033[38;5;1;48;5;220m Some text\033[m"
```
wheres in usual terminal (say `iterm2`) it looks fine (red text on yellow background).

Note, that the individual color changing commands, e.g:
```
 echo -e "\033[38;5;1m Some text\033[m"
```
and
```
echo -e "\033[48;5;220m Some text\033[m"
```
works fine.

P.S. If it helps: `$TERM` is `xterm-256color`

cgull:
Fix merged.

cgull:
Sorry, @keithw, but I think this is the right thing to do.  Terminfo/termcap were never intended as complete descriptions of terminal behavior; they're just databases that curses/terminfo can refer to to figure out what they should do.  All of the other emulator code I've looked at (I think there were at least 3: xterm, gnome-terminal, rxvt maybe?) works this way.

timofonic:
@pankdm @cgull 

Is this finally solved? (sorry for asking in different issues, not sure where to ask...)

keithw:
The issue is closed because it is solved. If you use mosh and find a problem, please open a new issue.

PR #939, opened 2017-10-26
Add true color support.

kangjianbin:
Implement true color support define in:
https://en.wikipedia.org/wiki/ANSI_escape_code

The sequence is:
ESC[ … 38;2;;; … m Select RGB foreground color
ESC[ … 48;2;;; … m Select RGB background color

### DIFF
--- a/src/terminal/terminalframebuffer.h
+++ b/src/terminal/terminalframebuffer.h
@@ -55,7 +55,6 @@ namespace Terminal {
   public:
     typedef enum { bold, faint, italic, underlined, blink, inverse, invisible, SIZE } attribute_type;
 
-    // all together, a 32 bit word now...
     static const unsigned int true_color_mask = 0x80000000;
     unsigned int foreground_color;
     unsigned int background_color;
--- a/src/terminal/terminalframebuffer.h
+++ b/src/terminal/terminalframebuffer.h
@@ -56,8 +56,9 @@ namespace Terminal {
     typedef enum { bold, faint, italic, underlined, blink, inverse, invisible, SIZE } attribute_type;
 
     // all together, a 32 bit word now...
-    unsigned int foreground_color : 12;
-    unsigned int background_color : 12;
+    static const unsigned int true_color_mask = 0x80000000;
+    unsigned int foreground_color;
+    unsigned int background_color;
   private:
     unsigned int attributes : 8;
 
@@ -67,6 +68,14 @@ namespace Terminal {
     void set_background_color( int num );
     void set_rendition( color_type num );
     std::string sgr( void ) const;
+
+    static unsigned int make_true_color( unsigned int r, unsigned int g, unsigned int b ) {
+      return true_color_mask | (r << 16) | (g << 8) | b;
+    }
+
+    static bool is_true_color(unsigned int color) {
+      return (color & true_color_mask) != 0;
+    }
 
     bool operator==( const Renditions &x ) const
     {
--- a/src/terminal/terminalframebuffer.cc
+++ b/src/terminal/terminalframebuffer.cc
@@ -505,6 +505,8 @@ void Renditions::set_foreground_color( int num )
 {
   if ( (0 <= num) && (num <= 255) ) {
     foreground_color = 30 + num;
+  } else if ( is_true_color( num ) ) {
+    foreground_color = num;
   }
 }
 
@@ -512,6 +514,8 @@ void Renditions::set_background_color( int num )
 {
   if ( (0 <= num) && (num <= 255) ) {
     background_color = 40 + num;
+  } else if ( is_true_color( num ) ) {
+    background_color = num;
   }
 }
 
@@ -544,13 +548,27 @@ std::string Renditions::sgr( void ) const
 
   ret.append( "m" );
 
-  if ( foreground_color > 37 ) { /* use 256-color set */
+  if ( is_true_color( foreground_color ) ) {
+    char col[64];
+    snprintf( col, 64, "\033[38;2;%d;%d;%dm",
+              (foreground_color >> 16) & 0xff,
+              (foreground_color >> 8) & 0xff,
+              foreground_color & 0xff);
+    ret.append( col );
+  } else if ( foreground_color > 37 ) { /* use 256-color set */
     char col[ 64 ];
     snprintf( col, 64, "\033[38;5;%dm", foreground_color - 30 );
     ret.append( col );
   }
 
-  if ( background_color > 47 ) { /* use 256-color set */
+  if ( is_true_color( background_color ) ) {
+    char col[64];
+    snprintf( col, 64, "\033[48;2;%d;%d;%dm",
+              (background_color >> 16) & 0xff,
+              (background_color >> 8) & 0xff,
+              background_color & 0xff);
+    ret.append( col );
+  } else if ( background_color > 47 ) { /* use 256-color set */
     char col[ 64 ];
     snprintf( col, 64, "\033[48;5;%dm", background_color - 40 );
     ret.append( col );
--- a/src/terminal/terminalfunctions.cc
+++ b/src/terminal/terminalfunctions.cc
@@ -424,6 +424,25 @@ static void CSI_SGR( Framebuffer *fb, Dispatcher *dispatch )
       i += 2;
       continue;
     }
+
+    /* True color support: ESC[ ... [34]8;2;<r>;<g>;<b> ... m */
+    if ( (rendition == 38 || rendition == 48) &&
+         (dispatch->param_count() - i >= 5) &&
+         (dispatch->getparam( i+1, -1 ) == 2)) {
+      unsigned int red = dispatch->getparam(i+2, 0);
+      unsigned int green = dispatch->getparam(i+3, 0);
+      unsigned int blue = dispatch->getparam(i+4, 0);
+      unsigned int color;
+
+      color = Renditions::make_true_color( red, green, blue );
+
+      (rendition == 38) ?
+        fb->ds.set_foreground_color( color ) :
+        fb->ds.set_background_color( color );
+      i += 4;
+      continue;
+    }
+
     fb->ds.add_rendition( rendition );
   }
 }
--- a/src/tests/emulation-attributes-truecolor.test
+++ b/src/tests/emulation-attributes-truecolor.test
@@ -1,0 +1,1 @@
+emulation-attributes.test
--- a/src/tests/emulation-attributes.test
+++ b/src/tests/emulation-attributes.test
@@ -1,9 +1,10 @@
 #!/bin/sh
 
 #
-# This validates VT100, 16-color, and 256-color attributes against
-# tmux.  It is not run directly, but as subtests based on the
-# executable's name for vt100, 16color, 256color8, and 256color248.
+# This validates VT100, 16-color, 256-color and true color attributes
+# against tmux.  It is not run directly, but as subtests based on the
+# executable's name for vt100, 16color, 256color8, 256color248 and
+# truecolor.
 # This is because Mosh internally represents the first 8 values of the
 # 256color space as though they were the 16-color values they are
 # equivalent to.  tmux does not filter this out on its redisplay, so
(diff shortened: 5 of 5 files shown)


eminence:
i did an initial test with minttty as a terminal (which supports true color), and things seem to work.  i will test some more later.  @kangjianbin can you say anything about scenarios or apps that you've already tested?

kangjianbin:
I only tested it with a patched Emacs. 

In windows, tested mosh in Mintty.

In Linux, tested it with xterm(which supports true color sequence but translate it to 256 colors) and xfce4-terminal.

src/terminal/terminalframebuffer.h
```diff
@@ -56,8 +56,9 @@ namespace Terminal {
     typedef enum { bold, faint, italic, underlined, blink, inverse, invisible, SIZE } attribute_type;
 
     // all together, a 32 bit word now...
```
andersk: Not a 32 bit word anymore.

andersk:
What happens on terminals that don’t support true color? Unless there’s a way to detect or configure that, we need to also preserve the 16 or 256 colors as a fallback.

kangjianbin:
Looks like there is no standard way to detect if a terminal supports true color yet (let me know if it is incorrect).
There are several combinations:
1. App is true color
      + Mosh supports true color and Terminal supports true color
                Everything is good
      + Mosh supports true color and Terminal doesn't supports true color
                Color is incorrect. 
     + Mosh doesn't support true color and Terminal supports true color
                Color is incorrect.
      + Mosh doesn't support true color and Terminal doesn't support true color
                Color isn incorrect.

2. App isn't true color
         Everything is good.

So in all cases, we dont' bring any regression but get benefit if the terminal supports TC.

We can also provide some configuration to let user choose if they want true color or not (for example, tmux). If TC is disabled, we fallback to 256 or 16 color mode.
But in my opinion, it is application's responsibility to decide if outputs TC sequence or not. If application chooses to output TC sequence, it is high possibility that the terminal supports TC or user forces it to output TC seq (like me :-)  ). In both cases, mosh needn't do the fallback.

andersk:
> 1. App is true color
>    * Mosh doesn't support true color and Terminal doesn't support true color
>     Color isn incorrect.

That’s where I’m skeptical. If the app is written well, then it will send both a reasonable 16/256 color fallback sequence and a true color sequence. The terminal ignores the true color sequence and you get the reasonable fallback color. By naïvely adding true color support to Mosh and not the terminal, you lose that reasonable fallback.

kangjianbin:
AFAIK, the behavior is different. I only verified Putty's behavior. Not easy to find a term that doesn't support true color in Linux.

For terminal (putty) that doesn't support true color, it parses the true color sequence to incorrect color instead of ignoring it. 
And looks like both emacs and vim send either true color sequence or 256 color, not both. (May be incorrect. I just analyzed the commint, didn't verify it).p

andersk:
The [colon-delimited codes](https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit) `\e[38:2:<r>:<g>:<b>m`, `\e[48:2:<r>:<g>:<b>m` are correctly ignored by PuTTY.

kangjianbin:
This patch doesn't handle colon-delimited codes. 
Looks like mosh doesn't handle the codes either?
So it is not a problem.

>Also, even with an application that isn’t written well, because Mosh’s screen-diffing optimizations may result in things being drawn in a different order, not being careful about this could result in not just wrong colors, but inconsistently changing wrong colors.

Would you please explain why this patch can result in inconsistent wrong colors?
When mosh-server receives true color seq, mosh-client outputs the same seq too. So looks like there isn't any inconsistency.

cgull:
@kangjianbin, thank you for doing this!  I thought we needed more complicated code for storing 24-bit attributes in a `Cell`, but I think your implementation will work correctly.

My limited research says there's no good/uniform way to detect 24-bit color support in a terminal, so I agree with @kangjianbin on that point.

I think we should implement the colon-delimited codes, rather than discarding them.  But I also think that real usage of the colon-delimited codes is nonexistent, so either choice is OK.

@kangjianbin, have you tried any synthetic testing of your code, with scripts that generate 24-bit color sequences?  Have you verified that >1 attribute can be chained together in the same sequence?  Your code looks like it should be OK, but please test this.

Also, please write a test case for this code.  This should be fairly easy, just add another case to the code in `src/tests/emulation-attributes.test`, and add a link and name in `makefile.am`.

@andersk, I expect "reasonably-written" apps that do the 16/256 color fallback you describe to be rare.  Can you give examples of such apps?  One issue is that semicolon-delimited codes for 24-bit color will be misinterpreted sometimes into various other attributes by terminal emulators that don't support 24-bit color.

24-bit color support is fairly common now. There are major exceptions like the MacOS bundled `Terminal.app`, and some of the major Android apps, see (https://gist.github.com/XVilka/8346728).  It does appear that decent emulators with 24-bit support are now available with little effort on most platforms.  I think many uses of 24-bit color schemes will map into 256-color space badly, and users will have to disable them when using older terminals in any case, with or without this mapping.  I would be comfortable simply documenting that Mosh simply passes whatever colors the app generates, and letting users handle conflicts between newer apps and older terminals.

cgull:
Another way to look at this is, how many GitHub issues will we get with different types of 24-bit support?

* No 24-bit support at all:  See #649.  :)
* @kangjianbin's simple 24-bit support, with no mapping or support for separate 24-bit and 256-color attributes:  I think this would be a small number, mostly from users that have 256-color terminals and apps that use 24-bit color.
* A more complex 24-bit support, with mapping or support for 24-bit and 256-color attributes simultaneously:  I think this would be an equal or smaller number to the previous choice.

andersk:
I think I withdraw my objection after noticing that we always reset attributes (`\e[0m`) before outputting a true color sequence. That means if the color is wrong, at least it will be deterministically wrong.

egmontkob:
> Looks like there is no standard way to detect if a terminal supports true color yet (let me know if it is incorrect).

Nope, unfortunately this is still correct.

> If the app is written well, then it will send both a reasonable 16/256 color fallback sequence and a true color sequence.

I'm not aware of any app doing this. Mosh could start doing this, however:

Outputting semicolons as delimiters, combined with a terminal emulator that doesn't understand true colors, might result in misparsing the RGB prefix "2" as well the R, G, B values as other ANSI codes. E.g. the intended color #030405 _could_ end up being dim (2), italic (3), underlined (4) and blinking (5) in some terminal emulators. I don't know what various terminal emulators that don't support true colors do here.

Outputting colons as delimiters is supposed to overcome this problem in the long run; terminal emulators can reliably ignore future parameters that take subparameters. Colons, however, are not supported by many terminal emulators. In turn, apps all decide to output semicolons. As such, this feature remains pretty much unused.

cgull:
@andersk:  I didn't think to ask this before-- do you have/see a use case where either storing/passing both 256 and 24-bit color, or downmapping from 24-bit to 256 color, is useful/important?  (Does anybody else, for that matter?)

andersk:
Well in theory, it would be nice to move toward a world where one doesn’t have to separately configure every application to use 24-bit color. That can only realistically happen if either (a) we get terminfo entries for 24-bit color and wait for them to become available on most systems, or (b) it’s possible to use 24-bit color in a backwards compatible way. In world (a), we need Mosh’s 24-bit color support to be configurable like 256 color support is now, and in world (b), we need Mosh to pass both 256 and 24-bit color (using colons for the latter). But the evidence suggests that both (a) and (b) are hypotheticals, so I don’t know what to say.

“Dear future readers who are dreaming up the awesome new terminal extensions of tomorrow, *please* try to think about these issues before you try to get your extensions widely deployed! Semicolons? Seriously??”

Okay that probably isn’t going to help, but it made me feel better.

Perhaps we could try to use our influence as the world’s second most popular remote terminal protocol to push for one path or the other?

andersk:
I just found
https://gitlab.com/gnachman/iterm2/issues/5294
https://bugzilla.gnome.org/show_bug.cgi?id=754521
https://bugs.kde.org/show_bug.cgi?id=371919
It seems that, lacking cooperation from the ncurses maintainers, terminal maintainers are converging on an unofficial standard of setting `COLORTERM=truecolor` in the environment to indicate true color support. An environment variable isn’t great for this purpose because it’s inherited in terminals launched from other terminals, but I guess it’s better than nothing.

keithw:
Maybe we should just ask Ted Dickey (terminfo maintainer?) if we can get an attribute allocated for this...

egmontkob:
Thomas, not Ted.
A couple of folks have already asked him, no success. But you could give it another try. That would be by far the cleanest solution.
(I cannot understand why adding a few text lines is such a no-go for him. terminfo is not meant to describe the features he likes, it's meant to describe how terminal emulators work and what they're capable of, no matter if he likes those features or not.)

keithw:
Thanks, do you have a pointer to any of those previous discussions? Would like to understand the history here.

egmontkob:
Follow the links from andersk's recent comment, they take you to some discussion (maybe not all) on the mailing list. I guess there was a bit more, like in 2017-09 another request that was left unanswered. Search through the [bug-ncurses](http://lists.gnu.org/archive/html/bug-ncurses/) mailing list archive; unfortunately I don't have a list of all the relevant links.

andersk:
> An environment variable isn’t great for this purpose because it’s inherited in terminals launched from other terminals, but I guess it’s better than nothing.

(Example where that’s broken today: `M-x shell` or `M-x term` or `M-x ansi-term` in Emacs launched from gnome-terminal. Emacs could of course work around this by unsetting the `COLORTERM` environment variable in its children—but is it reasonable to demand that every terminal emulator does so if they know nothing about true color?)

kangjianbin:
@cgull As I know, xterm do the down mapping. I guess it needs support some old VGA cards.
Tmux also do the mapping if true color isn't enabled.

kangjianbin:
Is there any update on this? Any suggestion is welcome.

cgull:
@kangjianbin, sorry to not do this earlier.  I didn't spend much time on Mosh this week.  I just pushed your work, and some additional changes, to master.  Thank you, again!

I think we are likely to find some more issues that need to be resolved, but this is a good start.

kangjianbin:
Ok thanks. My experience on mosh is limited, but I would like to help if this commit causes problems.

agriffis:
I'm seeing what looks like a discrepancy between true color with ssh and mosh. You can see it in the screenshots below, notice the background of vim-airline at the bottom of the screen. It also affects visual selection which doesn't show in mosh but shows as a similar background in ssh.

mosh 1.3.0.171108.6fb4d0d on client and server
nvim v0.2.1-953-g3e502fd7d (appimage) with [jellybeans](https://github.com/nanotech/jellybeans.vim) and [vim-airline](https://github.com/vim-airline/vim-airline)

**MOSH**
screenshot 2017-11-08 at 10 55 35 pm

**SSH**
screenshot 2017-11-08 at 10 56 33 pm

agriffis:
Hmm, it seems to be related to the front-end terminal. Those screenshots are on a Chromebook running [Secure SSH](https://chrome.google.com/webstore/detail/secure-shell/pnhechapfaindjhompbnflcldabbghjo?hl=en) to localhost, then mosh to server. So the terminal emulator is [hterm](https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm)

If I mosh from gnome-terminal, then I don't see the problem.

So it appears that it's particular to running mosh in hterm. However true color works fine with OpenSSH, as evidenced by the screenshots. So what is mosh doing that hterm doesn't like?

agriffis:
Related conversation on the chromium-hterm mailing list: https://groups.google.com/a/chromium.org/forum/#!topic/chromium-hterm/pjbFAd3EVgI

cgull:
Yup, somewhat aggressive implementation of SGR sequence on my part, combined with a bug on hterm's part.  I'm glad to see hterm is picking up a fix so quickly.  I will leave mosh as-is for the moment, but I may change mosh to send separate SGR sequences for each true-color attribute before our next release, to eliminate potential compatibility problems.

@rpwoodbu, you'll obviously want to ensure hterm and mosh-client are compatible, the next time you update mosh sources for https://github.com/rpwoodbu/mosh-chrome.

vapier:
it'll be fixed in the next hterm/Secure Shell release:
https://chromium-review.googlesource.com/765111

i'd prefer to not hack mosh to accommodate bugs in hterm ... lets fix hterm! :)

cgull:
@vapier, agreed, but:
* Both hterm and mosh are upstreams for https://github.com/rpwoodbu/mosh-chrome, if we and @rpwoodbu do releases before you do, he'll have to patch mosh or hterm.
* I'm waiting to see if any other terminal emulators have problems with those SGR sequences.  (All of the ones I've tried so far are OK.) 

It's not a "hack" to fix it, it's just a minor deoptimization, anyway.

egmontkob:
It's damn unlikely that any other terminal emulator suffers from this exact same problem. It's much more likely that other apps also emit escape sequences that hterm used to misparse, e.g. [img.sh](https://git.gnome.org/browse/vte/tree/perf/img.sh?h=vte-0-48) does.

Whether to work it around in mosh sounds more like a philosophical than a technical question to me. Probably either working around straight away or leaving as-is is better than continue discussing it, see [xkcd: Efficiency](https://xkcd.com/1445/) :-)

andersk:
> It's damn unlikely that any other terminal emulator suffers from this exact same problem.

@egmontkob Mosh 1.2.4 and earlier do (for 256-color sequences, #507).

vapier:
i've just released hterm-1.75 which fixes things, so it should be all set on that side now

vapier:
for anyone else looking at img.sh, don't use the one linked above ... that older version (in the 0.48 branch) is broken and emits things like `\e[38;2;32795;20365;12392m`.  the [xterm doc](http://invisible-island.net/xterm/ctlseqs/ctlseqs.html) say:

> RGB values range from zero (0) to 255

if you look at a [newer version](https://git.gnome.org/browse/vte/tree/perf/img.sh?h=0.51.1), it renders fine in hterm.

egmontkob:
It used to work fine with imagemagick versions from those days, but then imagemagick changed its output format :P

Issue #942, opened 2017-11-17
mosh-server incorrectly reports detached session in utmp/wtmp


cgull:
In #559, @mojoTX reports that mosh-server reports sessions as detached when they aren't really.  I've verified this.  This is almost certainly fallout from the recent changes to add syslog.

mojoTX:
If I get some free time soon, I will take a look at your recent changes to `src/frontend/mosh-server.cc`, to see if I can identify the problem.

mojoTX:
@cgull and @amishHammer: It appears to have been a simple typo in commit `ab31b0f`.

mojoTX:
This is definitely fixed in pull request #944.

eminence:
@cgull  in #944 reported successful testing, so I think this issue can now be closed.

PR #944, opened 2017-11-21
Use HAVE_UTEMPTER instead of HAVE_UPTEMPTER

mojoTX:
Fix typo for issue #942:
- mosh-server incorrectly reports detached session in utmp/wtmp

### DIFF
--- a/src/frontend/mosh-server.cc
+++ b/src/frontend/mosh-server.cc
@@ -664,7 +664,7 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
   #ifdef HAVE_UTEMPTER
   bool connected_utmp = false;
   #endif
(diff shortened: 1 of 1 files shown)


mojoTX:
I redid the pull request since my original pull request inadvertently had the wrong email address in the commit.

cgull:
Thanks for the fix!  Sorry for being slow to take it-- I'd actually discovered and fixed it, but not gotten to some testing I wanted to do.  I've now checked all 4 combinations of `--enable-syslog` and `--with-utempter` and it all seems to be OK.

mojoTX:
No worries. :-) 

I was getting tired of seeing the error message, so decided to dig into it myself. It was pretty easy to spot (and fix) once I started looking through the recent commits in Git.

PR #947, opened 2017-12-03
mosh.pl: Allow shell expansion of --server with --local

### LINKED ISSUE #946: --local can't be used with --server if it contains spaces
As of mosh 1.3.2, this works:

```
$ mosh --server='env foo=bar mosh-server' 127.0.0.1
```

But this doesn't:

```
$ mosh --local --server='env foo=bar mosh-server' 127.0.0.1
Can't exec "env foo=bar mosh-server": No such file or directory at /usr/bin/mosh line 390.
Cannot exec env foo=bar mosh-server: No such file or directory
/usr/bin/mosh: Did not find mosh server startup message. (Have you installed mosh on your server?)
```

The error message points to line 390: `exec( $server, @server );`

In the nonlocal case, `$server` is not used as the program filename, but as part of the shell command that runs on the remote side: `my @exec_argv = ( `(...)`, $ssh_connection . "$server " . shell_quote( @server ) );`

I don't know Perl, but would something like this work? `exec( "$server " . shell_quote( @server ) );`

andersk:
Fixes #946, matching the behavior of --server without --local.

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -396,7 +396,7 @@ if ( $pid == 0 ) { # child
     delete $ENV{ 'SSH_CONNECTION' };
     chdir; # $HOME
     print "MOSH IP ${userhost}\n";
-    exec( $server, @server );
+    exec( "$server " . shell_quote( @server ) );
     die "Cannot exec $server: $!\n";
   }
   if ( $use_remote_ip eq 'proxy' ) {


scripts/mosh.pl
```diff
@@ -396,7 +396,7 @@ sub predict_check {
     delete $ENV{ 'SSH_CONNECTION' };
     chdir; # $HOME
     print "MOSH IP ${userhost}\n";
-    exec( $server, @server );
```
cgull: I'm wondering, is there any reason to use `shell_quote()` here at all?  We could just do `exec( "server " . join( " ", @server )`, since we're not running ssh here.

andersk: Joining an array with spaces and no quoting is *always the wrong thing to do*. It’s a fundamentally lossy operation. In this case, it would manifest by corrupting `@command`: `mosh --local ::1 -- cmd 'arg with spaces'` would result in `cmd arg with spaces`. Even if that bug weren’t as easy to manifest as it is, it would still be locally a bug. Don’t do it. *Don’t dooo ittt.*

TomiBelan: Looks good to me (unsurprisingly...), though as I mentioned I didn't test it myself.

cgull:
@TomiBelan: I still want to know what your use case is-- is it just devel/experimental use of Mosh, or have you got some weird situation where `mosh --local` helps you use Mosh where `mosh localhost` wouldn't work?

TomiBelan:
Oh, I misunderstood and thought you want my response to this PR. I commented in #946.

scripts/mosh.pl
cgull: Actually, looking at this again, this code does break if there's a separator or metacharacter in the executable pathname (such as `/Users/cgull/spacey directory/mosh/src/frontend/mosh-server`).  I believe it should be `exec( shell_quote( $server, @server ) )`, so as to quote the executable name too.  I *think* I'm understanding Perl correctly here.  @andersk, concur?

TomiBelan: That doesn't sound right. If the goal is to make `--server='env foo=bar mosh-server'` work with and without `--local`, then spaces in `$server` should *not* be quoted.

andersk: Right, like @TomiBelan said, `--server` is intentionally passed without quoting so the shell can split it (19884d6c85c15b44b10c23c42d5687d1f9667bcd), and the point of this PR is to fix `--local` to match that behavior.

cgull: I started looking at a shell quoting problem in `src/tests` that this interacts with and forgot about the original problem...

Pulled now.

PR #954, opened 2018-01-18
Type Select::got_signal as volatile sig_atomic_t

### LINKED ISSUE #634: Fix signal handler
[The variable "got_signal"](https://github.com/keithw/mosh/blob/c58642dd63b8d359a44bdf1197df7b8405e25d94/src/util/select.h#L207) should [use the data type "sig_atomic_t"](https://www.securecoding.cert.org/confluence/display/c/SIG31-C.+Do+not+access+shared+objects+in+signal+handlers).
I guess that a different program design will be needed for your function "[handle_signal](https://github.com/keithw/mosh/blob/f3073bbe5dafb892ce0c1c93dd246233eaa73309/src/util/select.cc#L39)".

andersk:
Fixes #634.

### DIFF
--- a/src/util/select.h
+++ b/src/util/select.h
@@ -74,7 +74,11 @@ private:
 
   void clear_got_signal( void )
   {
-    memset( got_signal, 0, sizeof( got_signal ) );
+    for ( volatile sig_atomic_t *p = got_signal;
+          p < got_signal + sizeof( got_signal ) / sizeof( *got_signal );
+          p++ ) {
+      *p = 0;
+    }
   }
 
   /* not implemented */
@@ -227,9 +231,9 @@ private:
 
   int max_fd;
 
-  /* We assume writes to these ints are atomic, though we also try to mask out
+  /* We assume writes to got_signal are atomic, though we also try to mask out
      concurrent signal handlers. */
-  int got_signal[ MAX_SIGNAL_NUMBER + 1 ];
+  volatile sig_atomic_t got_signal[ MAX_SIGNAL_NUMBER + 1 ];
 
   fd_set all_fds, read_fds;
 


eminence:
Error from travis:

> mosh-client: ../../../src/frontend/../util/[select.h:195](https://github.com/andersk/mosh/blob/36e7394aa51c0375d9722bc4a5f0e25e609e69f4/src/util/select.h#L195): bool Select::read(int) const: Assertion `((((&all_fds)->fds_bits)[\_\_extension__ ({ long int __d = (fd); (__builtin_constant_p (__d) ? (0 <= __d && __d < 1024 ? (__d / (8 * (int) sizeof (__fd_mask))) : __fdelt_warn (__d)) : __fdelt_chk (__d)); })] & ((__fd_mask) 1 << ((fd) % (8 * (int) sizeof (__fd_mask))))) != 0)' failed.
> inpty: child exited with signal 6
> Aborted (core dumped)

andersk:
Er, right. It was late and I had temporarily forgotten how to write code. Fixed.

src/util/select.h
```diff
@@ -227,9 +231,9 @@ class Select {
 
   int max_fd;
 
-  /* We assume writes to these ints are atomic, though we also try to mask out
+  /* We assume writes to got_signal are atomic, though we also try to mask out
      concurrent signal handlers. */
-  int got_signal[ MAX_SIGNAL_NUMBER + 1 ];
+  volatile sig_atomic_t got_signal[ MAX_SIGNAL_NUMBER + 1 ];
```
elfring: The following implementation details cause corresponding development concerns for safe signal handling.
* Class attribute
* Usage of an array

My bug report is therefore not fixed so far.

By the way:
The property “`volatile`” could be provided by [the data type definition “`sig_atomic_t`”](https://bytes.com/topic/c/answers/220980-volatile-sig_atomic_t-redundant "Is \"volatile sig_atomic_t\" redundant?") already.

andersk: > The following implementation details cause corresponding development concerns for safe signal handling.
> * Class attribute
> * Usage of an array

Although previous versions of the C++ standard may have lacked the clarity to guarantee the safety of these extra reads, it’s guaranteed in C++14 after the integration of [N3910](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3910.html), which in particular adds the statement “A call to the function `signal` synchronizes with any resulting invocation of the signal handler so installed.”

> By the way:
> The property “`volatile`” could be provided by [the data type definition “`sig_atomic_t`”](https://bytes.com/topic/c/answers/220980-volatile-sig_atomic_t-redundant "Is \"volatile sig_atomic_t\" redundant?") already.

I don’t know who the commenters on that forum are, but the C and C++ standards are clear that special status in the memory model is granted only to `volatile sig_atomic_t`. (The “possibly volatile-qualified” language quoted on that forum is indeed confusing, but it only appears in the C standard, and the interpretation that this language makes the `volatile` redundant is not even consistent with what the rest of the C standard says.)

elfring: >… the function `signal` synchronizes …

This information does not matter for the other remaining software development concerns.


>…is indeed confusing, …

Can you omit a qualifier if it was specified for a `typedef` already?

andersk: > This information does not matter for the other remaining software development concerns.

I claim it addresses all the relevant concerns including the ones you have listed, so please be specific if you claim otherwise. The class attribute access and the array indexing are read operations that would have the desired effect at the point where the signal handler is installed, so the language I quoted guarantees that they have the desired effect within the invocation of the signal handler.

> Can you omit a qualifier if it was specified for a `typedef` already?

Yes, but no part of either standard suggests that `sig_atomic_t` is guaranteed to be already `volatile`-qualified, and indeed it is not `volatile`-qualified in at least glibc (it’s just a `typedef` for `int`).

elfring: >I claim it addresses all the relevant concerns including the ones you have listed, …

I disagree to your view partly.


>The class attribute access and the array indexing are read operations that would have the desired effect at the point where the signal handler is installed, …

I suggest to treat class member (together with array elements) accesses and the installation of a callback function as separate issues.


>Yes,

Thanks for a bit of consensus there.


>but no part of either standard suggests that `sig_atomic_t` is guaranteed to be already `volatile`-qualified, …

A specific implementation of the C/C++ standard specifications can choose which `typedef`is finally provided for a specific compilation configuration.

andersk: I’m not bringing up the installation of the signal hander as a separate issue—I’m bringing it up precisely to explain why the accesses you’re complaining about are safe. Your repeated assertions that the installation point is irrelevant do not make it irrelevant. It is relevant because the standard guarantees that it synchronizes with the handler containing the accesses in question. We can’t have a useful discussion if you aren’t going to take the time to understand the explanation. (Which is fine, because I don’t think there’s anything left to be said.)

I may have misunderstood your concern about the `volatile` qualifier. I thought you were suggesting (like some commenters in the bytes.com link you provided) that `volatile` was perhaps *unnecessary* because `sig_atomic_t` may already include `volatile`. I now see that you are suggesting that it’s perhaps *illegal* in such a case. The answer remains no: it is always legal and potentially necessary, so it stays.

elfring: >It is relevant because the standard guarantees that it synchronizes with the handler containing the accesses in question.

We interpret the provided data synchronisation in different ways.

Another approach for a better common understanding:
* Does the C language standard specification allow you store a `sig_atomic_t` variable in a class?
* From which source do you get the information that array accesses are async-signal-safe?


>I now see that you are suggesting that it’s perhaps illegal in such a case.

You interpret my feedback in an unwanted direction here.


If you would disagree to implementation details (like to the exact type definition from a known implementation of the discussed software standard), you can eventually adjust it.

andersk: > * Does the C language standard specification allow you to store a `sig_atomic_t` variable in a class?
> * From which source do you get the information that array accesses are async-signal-safe?

Mosh is C++ (and C doesn’t have classes), so I’ll assume that’s what you meant. I’ll further assume that you are asking whether it’s safe to access such a variable from a signal handler (and not merely to have such a variable at all, which of course is fine). The general answer to both questions, then, are “it depends”. The current C++ standard does not have a blanket prohibition against accessing signal handlers touching static objects that aren’t `volatile sig_atomic_t` like C99 and C++11 did. Instead, it carefully defines the circumstances when accesses may or may not result in a data race and prohibits data races. The definition is based on relations like “synchronizes with” and “dependency-ordered before” and “sequenced before” and “happens before” and so on. (The spifiness of `volatile sig_atomic_t` is encoded as one of many statements about the “happens before” relation.) The gist is that as long as there’s an appropriate chain of these relations to some access from any relevant stores, then there isn’t a data race and the access is fine. Read the standard if you want to understand the details—it’s complicated for a reason.

> You interpret my feedback in an unwanted direction here.

Uh, then what *are* you suggesting about the `volatile` qualifier? Do you disagree that the `volatile` I wrote is legal and necessary? Be specific. I’m fast growing tired of this conversation.

elfring: >…, so I’ll assume that’s what you meant.

* A `sig_atomic_t` variable needs to work also (and only?) without class support.
* Under which circumstances [could array accesses be async-signal-safe](https://stackoverflow.com/questions/9837343/difference-between-thread-safe-and-async-signal-safe "Difference between thread safe and async-signal safe")?


>Be specific.

[You mentioned an information](https://github.com/mobile-shell/mosh/pull/954#discussion_r162778634 "Is \"volatile sig_atomic_t\" redundant?") which tried to point out:

>>… that `volatile` was perhaps unnecessary because `sig_atomic_t` may already include `volatile`.

andersk: We’re not having a productive conversation here.

The goal of this PR is to make Mosh a correct program. Mosh is a C++ program, not a C program, so its correctness is to be judged by the C++ standard, not the C standard. The current C++ standard says that a class access, an array access, or any other kind of access from a signal handler is correct if its execution on the abstract machine defined by the standard does not contain a data race as defined by the standard. (That is different from what the current C standard says, although I expect the C standard will eventually be updated, but again, the C standard has nothing to say about Mosh.) If you think there’s a problem with this code, please [read the C++ standard](https://isocpp.org/std/the-standard) and point out where you think the data race is. I don’t expect this to be easy (indeed, I expect it to be impossible, for the reasons I’ve explained), but it’s clear that there is no other way for this conversation to continue productively. In particular, you are not adding anything to the conversation by linking to random forum posts that aren’t about C++ and predate the modern formulation of the C++ memory model.

It does not matter that some implementations may already include `volatile` in `sig_atomic_t`. As long as there exists any implementations that doesn’t (which, as I’ve said, is indeed the case), or even as long as there may exist such an implementation later, it is necessary for Mosh to specify `volatile` explicitly. If you want to make a further response here, please include a clear explanation what you claim should be changed in Mosh and why.

elfring: >We’re not having a productive conversation here.

I find our dialogue still constructive. Clarifications of the mentioned technical details have got a tendency to get longer.


>The goal of this PR is to make Mosh a correct program.

This is generally fine.


>…, so its correctness is to be judged by the C++ standard, not the C standard.

This should be also fine.


>The current C++ standard says that a class access, an array access, or any other kind of access from a signal handler is correct if its execution on the abstract machine defined by the standard does not contain a data race as defined by the standard.

Would we like to check the existance (or absence) of data synchronisation difficulties there anyhow?


>If you think there’s a problem with this code, please read the C++ standard and point out where you think the data race is.

I imagine that class member and array accesses trigger some challenges for async-signal-safety.

Which versions would you like to support from this programming language standard?



>I don’t expect this to be easy …,

My view is similar in this case.


>but it’s clear that there is no other way for this conversation to continue productively.

It seems that we come closer in small steps, don't we?


>… and predate the modern formulation of the C++ memory model.

* In which software releases can you depend on this evolution already?
* [Where does the configuration script check](https://github.com/mobile-shell/mosh/blob/ab31b0f2711a11b53b22db8181713fe48cec0ce9/configure.ac#L12 "build specification") for this functionality as an essential build requirement?


>It does not matter that some implementations may already include `volatile` in `sig_atomic_t`.

I got an other development view.


>As long as there exists any implementations that doesn’t …,

The qualifier was omitted for a reason in a specific compilation configuration then.


>…, it is necessary for Mosh to specify volatile explicitly.

* This can be your software design decision.
  * The commit and pull request description should contain a bit more information than just the hint “Fixes #…”.
  * Would you like to annotate it any further?
* How much do you trust the data type definitions which you got from a selected software implementation?


>If you want to make a further response here, please include a clear explanation what you claim should be changed in Mosh and why.

How do you think about other software design options?
* [sigwaitinfo()](http://pubs.opengroup.org/onlinepubs/9699919799/functions/sigwaitinfo.html "Wait for queued signals")
* [Self-pipe trick](http://cr.yp.to/docs/selfpipe.html "The self-pipe trick")
* [Signal-safe locks](https://locklessinc.com/articles/signalsafe_locks/ "Description for signal-safe locks")

andersk: > Would we like to check the existance (or absence) of data synchronisation difficulties there anyhow?

What do you mean, “would we like to check”? Of course we would like to check. Either there is a data race or there isn’t a data race. I’m telling you that I checked, and there isn’t a data race, and I explained why. If you disagree, point out where the data race is. (Remember that “data race” is a technical term with a specific definition in the standard. It is not a generic synonym of “data synchronisation difficulty”. Again, if you are unwilling to read the standard and communicate using the language that it defines, this conversation cannot move forward.)

> I imagine that class member and array accesses trigger some challenges for async-signal-safety.

We’re not playing a game of imagination. I’m telling you that these accesses do not cause a data race in this code, and I explained why. If you disagree, point out where the data race is.

> Which versions would you like to support from this programming language standard?
> In which software releases can you depend on this evolution already?

The C++ memory model was constructed to codify the guarantees already provided by existing implementations, not the other way around. We can depend on all compilers to satisfy these guarantees, inasmuch as we can depend on compilers to do anything. Is this unsatisfying? Yes. Such is the nature of C++ development.

> The qualifier was omitted for a reason in a specific compilation configuration then.

The qualifier was omitted because the standard places the burden on the programs that use `volatile sig_atomic_t`, not on the implementations that provide the `sig_atomic_t` type, to write the qualifier if they rely on the associated guarantees.

elfring: >We can depend on all compilers to satisfy these guarantees, inasmuch as we can depend on compilers to do anything.

* You can configure this development tool for the programming language standard which you would like to require.
* Do you care for [system configuration checks](https://github.com/mobile-shell/mosh/blob/ab31b0f2711a11b53b22db8181713fe48cec0ce9/configure.ac#L12 "Checking for software functionality") around available standard versions?

PR #960, opened 2018-02-16
Improve(?) terminfo handling

### LINKED ISSUE #959: xterm-termite: unknown terminal type
I'm using the [termite terminal](https://github.com/thestinger/termite) and want to start a new instance with mosh running in it.

I'm starting it like this:

```
termite --hold -e 'mosh some.server'
(termite:10041): GLib-WARNING **: GChildWatchSource: Exit status of a child process was requested but ECHILD was received by waitpid(). Most likely the process is ignoring SIGCHLD, or some other thread is invoking waitpid() with a nonpositive first argument; either behavior can break applications that use g_child_watch_add()/g_spawn_sync() either directly or indirectly.
```

And the new terminal shows this mosh error:
```
'xterm-termite': unknown terminal type.
Died at /usr/bin/mosh line 311.
```

mosh shoul gracefully catch an unknown terminal type and not fail imo.

cgull:
Fixes #959, or something.

This makes Mosh more generous in the face of an unknown `TERM` variable or otherwise bad terminfo configuration.  It allows Mosh to stumble onward, disabling all terminfo-driven features except 256-color mode.

I am somewhat divided about this "fix".  On the one hand, it'll break quite horribly on a non-ANSI terminal such as an ADM-3a or Televideo 970.  On the other hand, terminals that are truly incompatible with Mosh are rare now, and this might improve the user experience.  Anybody have strong feelings one way or the other?

### DIFF
--- a/src/frontend/mosh-client.cc
+++ b/src/frontend/mosh-client.cc
@@ -89,13 +89,20 @@ static void print_usage( FILE *file, const char *argv0 )
 static void print_colorcount( void )
 {
   /* check colors */
-  setupterm((char *)0, 1, (int *)0);
-
-  char colors_name[] = "colors";
-  int color_val = tigetnum( colors_name );
-  if ( color_val == -2 ) {
-    fprintf( stderr, "Invalid terminfo numeric capability: %s\n",
-	     colors_name );
+  int errret = -2;
+  int ret = setupterm((char *)0, 2, &errret);
+  int color_val = 256;
+  if ( ret != OK ) {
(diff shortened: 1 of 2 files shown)


keithw:
I guess my tentative view is that if the user's setup is clearly messed up, it's preferable to give the user a clear error message at startup time (one we can help them with, or they can Google, or ideally one that tells them specifically what's wrong), than to silently ignore the problem and end up giving bad results in the middle of the session, which are much harder to debug and more likely to make Mosh seem flaky...

cgull:
Overcome by lack of interest.

Issue #120, opened 2012-04-10
SSH Agent Forwarding


dt:
Would be great to be able to forward local SSH agents. Seems to ignore `ForwardAgent yes` in `~/.ssh/config`

termie:
+1

xtimmu:
+1

NoUseFreak:
:thumbsup:

benlemasurier:
+1

timesking:
+1

jravetch:
:+1:

takaidohigasi:
+1

RubenKelevra:
+1

codeshrew:
:+1:

aaroncline:
:+1:

PAStheLoD:
Hey,

Since the code works (thanks to kevinr), I couldn't stand it just sitting somewhere in a lonely Git repo. So, here's a Debian (actually Ubuntu) package, that works (for me at least): https://github.com/PAStheLoD/mosh/tree/master#but-if-you-do

joshSzep:
+1

dejadejoder:
I may have founded a workaround using 'tmux':
1. login with normal ssh client
2. open a tmux terminal and detach it
3. login with mosh
4. attach previously created tmux terminal
You may be able to log with ssh-agent to your remote machines. I am not sure how long tmux will keep link with the ssh-agent...

patcon:
Just as one sample use-case that would seem to not work with the `ProxyCommand` solution:

When working in a country with poor internet connection, and wanting to treat your local workstation as a stupid machine that SSH's into your actual workstation on a cloud provider. Using agent forwarding seems to make it as if my SSH key were deleted each time I log out

ddarbyson:
+1

davydovmax:
Must have feature for me (can't use mosh without it).

rinne:
The current version of agent patches on top of the current upstream master branch of mosh is in #696. Some maintainer asked for patches to be re-merged and I assumed that it meant that they were going to pull the feature to mainline. However it has been 17 days (after I re-merged the feature) and nothing has happened since.
For now, cloning mosh sources from https://github.com/rinne/mosh and compiling agent-forwarding-merge-20151128 branch (--with-agent-forwarding), does the trick.

Omeryl:
:+1:

gavit:
Has it been added to the main yet?

andr-04:
[...]

PR #986, opened 2018-07-11
Multiple network stream support

d4l3k:
This adds the basic support for multiple network streams via a `Network::MultiplexerStream`. It aggregates the diffs for multiple child streams into one.

This is intermediate work towards implementing #337.

### DIFF
--- a/src/protobufs/Makefile.am
+++ b/src/protobufs/Makefile.am
@@ -1,4 +1,4 @@
-source = userinput.proto hostinput.proto transportinstruction.proto
+source = userinput.proto hostinput.proto transportinstruction.proto stream.proto
 
 AM_CPPFLAGS = $(protobuf_CFLAGS)
 AM_CXXFLAGS = $(WARNING_CXXFLAGS) $(HARDEN_CFLAGS) $(MISC_CXXFLAGS) -Wno-error


brainstorm:
Can someome @mobile-shell review this please? @cgull ?

andersk:
This didn’t pass the automated CI test, because we still support platforms with poor or missing C++11 support (that may [change](https://github.com/mobile-shell/mosh/pull/990) after this release, but for now it’s a requirement). When building on a C++11 system, `make check` shows a failing test. Also, some quick manual tests show this isn’t backwards compatible with the existing Mosh protocol in either direction.

d4l3k:
@andersk I realize the tests are failing, just wanted someone to give a thumbs up/thumbs down whether this approach would ever be accepted before fixing it up.

Does it need to be backwards compatible between versions? I can probably hack some support in if necessary.

keithw:
We haven't broken compatibility on the wire protocol since early 2012 and I'm not too enthusiastic to do it now -- so yeah, any protocol design here would probably need to be backwards compatible (or bundled with some huge major rev). You've probably seen the amount of anger we induced by turning down a _backwards-compatible_ protocol change that added support for ssh-agent forwarding, so I just want to be upfront that we're unlikely to take major protocol changes to the Mosh/SSP protocol that haven't been discussed and reached consensus on the mailing lists.

brainstorm:
Keith, thanks for your answer. I've quickly scanned the [mosh-devel ML archives](http://mailman.mit.edu/pipermail/mosh-devel/) up to the beginning of 2017 but couldn't easily find that thread/discussion... can you brief us on what would be an acceptable patch would look like if even backwards-compatible changes are (routinely?) rejected?

I know that the stakes are high for such a popular, stable and great project, but I just want to manage expectations on whether we'll likely see this feature landing the main codebase at all or we shouldn't even be trying at this point?

Perhaps there's some major rearchitecting going on for future major releases and this feature makes more sense there instead of trying to merge it now?

keithw:
Hello -- I appreciate your candor here. I was thinking of the discussions around https://github.com/mobile-shell/mosh/issues/120 (and on the mailing list, https://www.mail-archive.com/mosh-devel@mit.edu/msg00980.html).

I think something like this could be part of a "mosh v2" (and I also have my ideas about what might be nice to have in "mosh v2"), but we honestly haven't even started the process of talking about what might be in scope for that. Probably we might start by brainstorming what we really want to do here, and in what dimensions we want to improve on mosh v1, and then build the protocol around that. That would be my druthers for the way to go, but it's a big effort not yet begun... It's not like we've started to rearchitect.

brainstorm:
It's almost a year later now, is there a mosh v2 in sight where this feature could be included, @keithw? The bounty for this feature getting implemented keeps increasing: https://github.com/mobile-shell/mosh/issues/337#issuecomment-518970059 :)

PR #994, opened 2018-08-12
Added const to methods which 'ought to be const' based on Const Checker

eyolfson:
I double checked that everything compiled without issue. These methods were found to by our Const Checker research tool. Let me know if these methods should not be const. Thanks!

### DIFF
--- a/src/crypto/crypto.h
+++ b/src/crypto/crypto.h
@@ -97,7 +97,7 @@ namespace Crypto {
     Base64Key(PRNG &prng);
     Base64Key( string printable_key );
     string printable_key( void ) const;
-    unsigned char *data( void ) { return key; }
+    const unsigned char *data( void ) const { return key; }
   };
 
   class Nonce {
--- a/src/statesync/user.h
+++ b/src/statesync/user.h
@@ -89,7 +89,7 @@ namespace Network {
     void apply_string( const string &diff );
     bool operator==( const UserStream &x ) const { return actions == x.actions; }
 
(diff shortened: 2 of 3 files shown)


cgull:
We appreciate code-improvement PRs like this and generally take them if they're of decent quality, but I would make a few requests:

* Tell us what the tool is, and what it does.  (Google didn't turn up anything.)
* Add a target to Makefile.am so we understand what exactly was done, and so we can re-run the check when our code changes, or if the tool improves, later.
* Mosh's code is of fairly good quality but I'd be surprised if there were only four should-be-const methods in the code base, can you say how thorough the tool is in practice?
* In this case, a couple of the suggested changes seem to be ill-advised (or our code ill-designed): `reset_cell()` and `reset_row()` are always passed a pointer to a member of the object, and thus are clearly not const in practice, even if they appear to be so to a parser.

eyolfson:
It's an in-progress research tool. The goal is to eventually make it as usable as IWYU, but it's currently a prototype. Pending a submission it'll be released. This pull request is to see if developers would act on the suggestions. The code may need some more tuning (`reset_cell()` and `reset_row()`).

If you want I can remove the const on `reset_cell()` and `reset_row()` from the pull request, if you want to merge it. Thanks!

achernya:
Unfortunately this PR now has a lot of conflicts, and I believe the changes remaining are actually undesirable (the `reset_*` methods). I'm going to close this for now, please do re-submit a new version if your tool has been updated.

PR #1001, opened 2018-08-31
Improved unicode wcwidth support

cgull:
This is a revised version of #949.

Changes from #949 include:

* Implement East Asian Width narrow/wide support.  Automatically enable in C/J/K locales, and add `--eaw-wide` switch.
* Transmit and mostly store width tables as a list of runs, greatly reducing storage needed.  Also properly manage this width state in a class, rather than ad-hoc global state.
* Add an edit_widths.py script for merging/diffing width tables.
* Add a hold-session capability for a future day when the diff table is too large to fit in a single packet.
* The reference table is still Unicode 9, but the default table is now Unicode 11.  (The reference table will be set to Unicode 11 before being committed to Mosh.)
* General cleanup.

Notes:

I did a bunch of testing with various formats to see what was most space efficient.  Text proved more efficient than protobufs, lists of run lengths were better than lists of offsets.  The 9->11 overlay is 946 bytes uncompressed and 350 bytes gzipped standalone (actual compression will be something slightly different in Mosh packets).

Changing the character width tables is a change in the state of the terminal emulator.  I see two ways to handle that:  either set the tables before the user session is allowed to start, or fully integrate the character width tables into the terminal state machine.  I chose to set the tables at startup.  I think to fully integrate the state, we would have to send the overlay table to the server, which would then have to send it (or references to the changes) back to the client to keep it in sync.  Messy.

With a small overlay, nothing extra needs to be done to Mosh to hold back the user session-- a while ago, I added some functionality on the server to wait until the first message is received.  But if the overlay ever gets so large that we decide we need to send it in pieces in multiple messages, then we need to add some functionality for that.  That's what the hold-session commits are for.  I lean towards including this as currently-unused functionality so it's available for future versions of Mosh, but I'm not sure about that yet.


src/util/chwidth.cc
```diff
@@ -0,0 +1,135 @@
...
+  }
+  if (run_start != comparelen) {
+    diff << comparelen - run_start << ' ' << run_type << std::endl;
+  }
+  return diff.str();
+}
+
+int ChWidth::chwidth( wchar_t wc ) const
+{
+  if (static_cast<size_t>( wc ) > widths.size()) {
```
andersk: `>=`.

src/unicode/Makefile.am
```diff
@@ -0,0 +1,37 @@
...
+PYTHONPATH=$(PWD)/google-libapps/libdot/third_party/wcwidth/:$(PWD)
+
+UCD.zip:
+	curl -O https://www.unicode.org/Public/11.0.0/ucd/UCD.zip
+
+google-libapps/libdot/third_party/wcwidth/ranges.py:
+	git submodule init
+	git submodule update
+
+ucd/PropList.txt ucd/UnicodeData.txt ucd/EastAsianWidth.txt: UCD.zip
```
andersk: This target will run multiple times concurrently and race with itself when building with `-j`.

src/unicode/Makefile.am
```diff
@@ -0,0 +1,37 @@
...
+	mkdir -p ucd
+	unzip -o -d ucd UCD.zip
+	ln -fs ucd/PropList.txt ucd/UnicodeData.txt ucd/EastAsianWidth.txt .
+
+eaw_narrow_map eaw_wide_map: ucd/PropList.txt ucd/UnicodeData.txt ucd/EastAsianWidth.txt edit_widths.py make_widths.py google-libapps/libdot/third_party/wcwidth/ranges.py
```
andersk: And this one.

andersk:
> Transmit and mostly store width tables as a list of runs,

But they’re still expanded out to `vector<unsigned char>` at runtime. Did you decide that the extra 1.1 MB of memory per session is worth the performance improvement over binary search? That’s not necessarily an insignificant amount of memory to a phone or a virtualized server.

src/util/chwidth.cc
cgull: Thanks for catching that.

PR #1019, opened 2019-01-04
Apply latest consecutive resize, not earliest.

peadar:
If there are consecutive resize events in the userstream to be applied in
"serve", we should apply the last/latest one in the sequence, not the
first/earliest one.

This fixes a problem where a flurry of resize events (eg, generated
by a window manager resizing the client), can cause mosh to have an
out-of-date idea as to what the physical geometry of the window is.

### DIFF
--- a/src/frontend/mosh-server.cc
+++ b/src/frontend/mosh-server.cc
@@ -739,12 +739,11 @@ static void serve( int host_fd, Terminal::Complete &terminal, ServerConnection &
 	    const Parser::Action &action = us.get_action( i );
 	    if ( typeid( action ) == typeid( Parser::Resize ) ) {
 	      /* apply only the last consecutive Resize action */
-	      while ( i < us.size() - 1 ) {
-		const Parser::Action &next  = us.get_action( i + 1 );
-		if ( typeid( next ) != typeid( Parser::Resize ) ) {
-		  break;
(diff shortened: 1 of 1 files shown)


cgull:
Ouch, this is a little embarrassing, especially since there's a test for that in the build.

@peadar, exactly how did you provoke this bug?  I've not seen it, and I'd like to try and reproduce it and see if I can improve `src/tests/window-resize.test`.  (Or, be my guest and do so yourself.)

peadar:
I can look at doing an automated test, but I have a utility that moves windows around an X11 display from the command line, and that's how I noticed. I distilled a few pieces to make it easier to repro here: https://github.com/peadar/x11-resize-test (You'll need development libs for X11 to compile)

If you connect to a mosh server with a few milliseconds latency in one window, and from another build this, run "./test -p", and click on the mosh server window, it'll rapidly flap the size between 1/2 and 1/4 of the original, and then restore it. For me, the mosh server shell has the wrong window size at the end, as reported by 'stty -a'

andersk:
It turns out I run into this reliably by maximizing or unmaximizing GNOME Terminal (on GNOME Shell, Wayland, Ubuntu 19.04)—apparently that generates multiple resizes at once. I’ve confirmed this fix. I’m not sure if we’re waiting for better tests to merge this, but that can always be done later.

PR #1024, opened 2019-02-04
configure: Add --enable-static-LIBRARY options for selective static linking

andersk:
Statically linking `mosh-server` with glibc isn’t a great idea for various reasons (nsswitch modules, locale format incompatibilities). But we can provide most of the benefits of static linking by allowing specific library dependencies to be linked statically using `-Bstatic` and `-Bdynamic`.  The full set at this time is

    ./configure --enable-static-libstdc++ --enable-static-libgcc \
      --enable-static-utempter --enable-static-zlib --enable-static-curses \
      --enable-static-crypto --enable-static-protobuf

which results in binaries whose only runtime library dependencies are
provided with libc:

    $ ldd src/frontend/mosh-server
            linux-vdso.so.1 (0x00007ffe0b377000)
            libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007fa0d9970000)
            libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fa0d97e3000)
            libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fa0d97c2000)
            libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa0d95d8000)
            /lib64/ld-linux-x86-64.so.2 (0x00007fa0d9f6a000)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -285,7 +285,7 @@ PKG_CHECK_MODULES([TINFO], [tinfo],
         AC_SUBST([TINFO_LIBS], ["$TINFO_LDFLAGS -ltinfo"])],)])])
 
 # Then try to find it in a specific install dir
-AC_ARG_WITH(curses, AC_HELP_STRING([--with-curses=DIR], [Where curses is installed]),
+AC_ARG_WITH(curses, [AC_HELP_STRING([--with-curses=DIR], [Where curses is installed])],
     [if test $withval != yes; then
         cv_curses=$withval
     fi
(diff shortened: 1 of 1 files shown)


cgull:
This seems like a good idea, but I would consider adding a `--enable-static-libraries` umbrella option so that packagers and users don't need to manually track all of Mosh's dynamic dependencies.

andersk:
Sure. I’ve added `--enable-static-libraries`.

PR #1050, opened 2019-07-04
Use GPL v3 license verbatim [skip ci]

mbrukman:
Replaced with pristine copy of GPL v3 from the canonical
source: https://www.gnu.org/licenses/gpl-3.0.txt

The current license file appears to have minor differences
from the canonical GPL v3 license, which causes GitHub to
not recognize it and label it correctly as GPL v3.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,7 +1,7 @@
                     GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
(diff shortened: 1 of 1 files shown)


andersk:
The upstream text was edited in place long after Mosh adopted it:

https://web.archive.org/web/20170914030415/http://www.gnu.org/licenses/gpl-3.0.txt
https://web.archive.org/web/20170930072152/http://www.gnu.org/licenses/gpl-3.0.txt

and GitHub’s licensee library was updated accordingly here:

https://github.com/licensee/licensee/pull/317/files#diff-cb5cb76367ec37937600f38e3193acb1

Whether we accept the upstream edits or not, perhaps a bug should be filed with licensee about recognizing old versions.

mbrukman:
Per the comment https://github.com/licensee/licensee/issues/387#issuecomment-509289065 upstream, it looks like this file is not the issue, but it's the presence of other files in the repo with licenses that is causing confusion for the tool. I'll close this PR as it's unnecessary; sorry to bother you.

PR #1052, opened 2019-07-09
Move generated includes to their own directory to avoid conflicts.

### LINKED ISSUE #1051: mosh doesn't build with libc++ >=7.0 on case-insensitive filesystems
The Mosh build process writes a file called `VERSION` in its root directory. clang >= 7.0 has an include called `version`. On a case-insensitive filesystem these are the same name, so `VERSION` ends up getting parsed as an include file.

Possible fixes:
- Don't put the root directory in the include path (and move config.h/version.h into a subdirectory)
- Rename `VERSION` to something else (`VERSION.txt`?)
- Generate `version.h` directly and not generate `VERSION` at all
- File an upstream bug against `clang`, since this affects a bunch of packages that have `VERSION` files

cgull:
Fixes #1051, mosh fails to build on case-insensitive filesystems.

I think putting generated includes in a separate directory is the best strategy here to avoid having this come up again, but autoconf/automake do drop a couple of extension-less files in that directory (`Makefile`, `stamp-h1`).  Let's hope the C++ standard committee never adds includes with *those* names.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,7 @@
 ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = scripts src man conf
 EXTRA_DIST = autogen.sh ocb-license.html README.md COPYING.iOS
-BUILT_SOURCES = version.h
+BUILT_SOURCES = VERSION
 AM_DISTCHECK_CONFIGURE_FLAGS = --enable-compile-warnings=distcheck --enable-examples --enable-syslog
 
 # non-Automake defines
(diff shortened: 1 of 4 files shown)


quentinmit:
It sounds like `BUILT_SOURCES` has some limitations, specifically that "make mosh-client" will fail unless "make all" has been done first. https://www.gnu.org/software/automake/manual/html_node/Built-Sources-Example.html suggests some alternatives that don't have that problem. I dunno if that's worth changing, since it already exists on HEAD.

I think you can avoid having autoconf/automake generate any files in that directory if you [put the rules in the parent directory's Makefile.am instead](https://www.gnu.org/software/automake/manual/html_node/Alternative.html#Alternative) and reference the generated file as `nodist_INCLUDES=include/version.h` instead of `nodist_INCLUDES=version.h`

quentinmit:
Yeah, on second thought, I think it's preferable to keep the patch small, so ignore what I said about `BUILT_SOURCES`. It's probably still worth trying the suggestion from the second link, though.

cgull:
The make-sub-target issue is actually somewhat important, for platforms that might only need `mosh-server` and not want to deal with the client's Perl dependencies.  I hadn't thought of that until you gave the example, thanks.

Your suggestion about using the parent dir's Makefile: yeah, that'd build OK, but I think it doesn't actually *solve* the problem, because then Automake would be doing `-I src/` instead of `-I src/include/`, and you'd just be looking at the generated files *there* instead.   Plus, need to change to `#include <include/version.h>` which is a little odd in its own right.

quentinmit:
I don't think moving it to the parent directory's Makefile does anything to the `-I` arguments or the include path you need to use in your source files.

cgull:
On closer inspection, I think you're right.  autoconf/automake automagically generates that `-I src/include` for `config.h`, and moving the generation of `version.h` around wouldn't affect that.

The subset builds for mosh-client or mosh-server only aren't a problem; those are handled at configure time, not by selecting different make targets.

cgull:
Committed as 1219911.

PR #1053, opened 2019-07-10
If exec()ing the remote command fails, pause briefly

cgull:
This came up on IRC, where user concatime was asking why `mosh -p PORT NAME 'tmux attach -t irc'` didn't work.  (The correct command is `mosh -p PORT NAME -- tmux attach -t irc`).  I realized that it's hard to debug this because mosh-server prints an error message for the failed exec, but exits immediately, and normally mosh-client clears the screen on exit.

This adds a 3-second delay before exit, after `exec()` fails, so the error is visible.


Kriechi:
Thank you! I encountered the same issue soo many times! ❤️ 👍 🎉

Kriechi:
@andersk @cgull anything holding this PR back?

PR #1055, opened 2019-07-17
Name the server on timeout banner

### LINKED ISSUE #1049: Mosh timeout banner does not name the host(s) on the connection
Mosh's timeout banner at the top of the screen does not identify the host(s) involved.  (`mosh-client` doesn't ever say which client or server hosts are involved, I think.)  This can be momentarily confusing when you're using Mosh both into and out of a jump box, and one of the links dies-- you can't tell which Mosh session is offended at the situation.  This is a pretty minor issue, but it just happened to me.

tansly:
Fixes #1049.

Shows the server's IP address on the timeout banner.

I can improve this if necessary. A thing that crosses my mind:
- This still can be confusing for the user because the user may have started mosh using a domain name thus does not necessarily know the IP address of the server. For better UX, I may try to display what the user gave as an argument when starting mosh. But this may be a problem if the hostname is overly long and does not fit in the banner.


tansly:
Fixed errors in CI.

Kriechi:
@andersk @cgull anything holding this PR back?

tansly:
> Hi all, I'm new here. Started using blink on an iPad and fell in love with mosh. I've been using two server: a local one in TLV and one from my client in us-east-1 and mosh just works.
> 
> 
> 
> My two cents is that the code looks good, but it has no tests. Probably because this message is not covered in the current e2e tests. Could be a good time to add this test case.

Hi @daonb, I would work on it but the latest commit to this repo is about 10 months ago. I’m not sure that it is active right now. I’m holding this until there is some activity.

PR #1059, opened 2019-08-08
Add support for ECMA-48 escape sequences for faint/half-bright and crossed-out/strikethrough

stanspas:


Juerd:
What needs to be added or changed before this can be merged? It's a trivial change, it works, and it's a feature I'd really like to have.

I'm currently patching mosh on several machines by hand, because a program uses faint text, which mosh does not yet support. It would support it with this PR.

I've checked the resources, and the ANSI escape sequences used in these commits are correct, including the dual functionality of 22 which unsets both dim and bold.

Please consider merging this PR. If any further work is needed, I might be willing to do it. But from the looks of it, it looks very correct and complete.

carloscabanero:
Updated with new formatting here too: https://github.com/blinksh/mosh-server/tree/dim-and-strikethrough.

PR #1062, opened 2019-08-14
configure: fix ffs check with _GNU_SOURCE

neheb:
According to the man page, ffs requires this or others to be defined.
_GNU_SOURCE tends to be inclusive so use that for simplicity.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -484,7 +484,8 @@ AC_CHECK_DECL([mach_absolute_time],
 AC_CHECK_DECLS([__builtin_ctz])
 
 AC_CHECK_DECLS([ffs], [], [],
-  [[#include <strings.h>]])
+  [[#define _XOPEN_SOURCE 700
+#include <strings.h>]])
 
 AC_CHECK_DECLS([be64toh, betoh64, bswap64, __builtin_bswap64], [], [],
   [[#if defined(HAVE_ENDIAN_H)


achernya:
According to my copy of the `ffs(3)` manpage, `_GNU_SOURCE` is only needed for `ffsl` and `ffsll` on glibc 2.27 and older. For `ffs`, it's `_XOPEN_SOURCE >= 700`, or since glibc 2.19 (2014), it's just default. I believe this patch is unnecessary. If you have a system where this fails, please let us know and we can re-evaluate.

neheb:
@achernya sure. I'm using musl. See https://git.musl-libc.org/cgit/musl/tree/include/strings.h#n24

achernya:
Okay, could you update your PR to use `_XOPEN_SOURCE` instead?"

neheb:
Will do. Another way to fix is to check function inside library.

neheb:
Fixed.

neheb:
I don't think that's needed. _GNU_SOURCE is probably defined somewhere.

achernya:
It is not safe to assume that is the case. Please explicitly add it.

configure.ac
```diff
@@ -484,7 +484,8 @@ AC_CHECK_DECL([mach_absolute_time],
 AC_CHECK_DECLS([__builtin_ctz])
 
 AC_CHECK_DECLS([ffs], [], [],
-  [[#include <strings.h>]])
+  [[#define _XOPEN_SOURCE
```
andersk, line 487: This needs to have a value of at least `700` for glibc. The “default” features complicate the observation of this requirement, but for example, this does not compile:

```c
#define _POSIX_C_SOURCE 200809
#define _XOPEN_SOURCE
#include <strings.h>
int main() { return ffs(1); }
```

and this does:

```c
#define _POSIX_C_SOURCE 200809
#define _XOPEN_SOURCE 700
#include <strings.h>
int main() { return ffs(1); }
```

See also [Feature Test Macros](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html).

neheb: Fixed.

neheb:
> It is not safe to assume that is the case. Please explicitly add it.

that solves no problem anywhere. this PR does as I can clearly see config.log and whether or not ffs() compiles or not.

achernya:
It solves the problem of a confusing regression on some platforms when header includes get refactored later on. Please add it to the file where it is used.

PR #1054, opened 2019-07-17
Support OSC 52 clipboard updates from tmux & others

gwicke:
Be more liberal in which options mosh accepts in clipboard updates, even
if those options are not preserved across the connection at this point.
This includes clipboard updates emitted by tmux, which is very
frequently used with mosh.

This eliminates the need for work-arounds like
https://gist.github.com/yudai/95b20e3da66df1b066531997f982b57b, which
did not even work for me.

PR #1089, opened 2020-01-28
Support OSC 52 clipboard updates from tmux & others

andrewmzhang:
# Personal note

Merging gwicke's branch with the mosh origin since I want to use this feature. 

==============
Be more liberal in which options mosh accepts in clipboard updates, even
if those options are not preserved across the connection at this point.
This includes clipboard updates emitted by tmux, which is very
frequently used with mosh.

This eliminates the need for work-arounds like
https://gist.github.com/yudai/95b20e3da66df1b066531997f982b57b, which
did not even work for me.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,8 +31,8 @@ init:
 # Install needed build dependencies
 # 
 install:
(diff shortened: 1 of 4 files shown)


andrewmzhang:
Sorry wrong branch.

andrewmzhang:
I may have broken travis test to: https://github.com/mobile-shell/mosh/pull/1054. 
Temporarily reviving this to make the test pass

PR #1095, opened 2020-04-24
Also set datadir for bash-completion pkg-config query

xfix:
bash-completion >= 2.10 uses datadir for its prefix path. See https://github.com/scop/bash-completion/commit/0cc34e8de658bd11fa7f728eb9852c7e29d8d6d4.

Motivation: https://github.com/NixOS/nixpkgs/issues/85893

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -513,7 +513,8 @@ AS_IF([test "$install_completion" != no],
      [if test "$prefix" = "NONE"; then
         completions="`pkg-config --variable=completionsdir bash-completion`"
       else
-        completions="`pkg-config --define-variable=prefix=$prefix --variable=completionsdir bash-completion`"
+        # bash-completion >= 2.10 uses datadir instead of prefix
+        completions="`pkg-config --define-variable=prefix=$prefix --define-variable=datadir=$datadir --variable=completionsdir bash-completion`"
       fi],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -513,7 +513,8 @@ AS_IF([test "$install_completion" != no],
      [if test "$prefix" = "NONE"; then
         completions="`pkg-config --variable=completionsdir bash-completion`"
       else
-        completions="`pkg-config --define-variable=prefix=$prefix --variable=completionsdir bash-completion`"
+        # bash-completion >= 2.10 uses datadir instead of prefix
+        completions="`pkg-config --define-variable=prefix=$prefix --define-variable=datadir=$prefix --variable=completionsdir bash-completion`"
```
andersk, line 517: `$datadir` is normally `$prefix/share`, no?

xfix: Oh, yeah, right, this won't work.

xfix:
Yeah, right, there is `if test "$prefix" = "NONE"; then` test, hm...

xfix:
Okay, I don't understand autoconf, whatever I do it's ugly if I want to support all possible scenarios. I think I will close this pull request, actually.

jtojnar:
I think it is fine to ignore the prefix=NONE case since it did not set prefix previously either.

PR #1118, opened 2020-11-02
README.md updates

ray3du:
Hey team, I have recently being trying to use mosh but I must stay I found it difficult. I found mosh on github and I felt that the lack of a how to get started help on the repository pushed me away from using it at first. I hope this commit will help new users with getting started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,25 @@ Other features
    * Mosh leverages SSH to set up the connection and authenticate
      users. Mosh does not contain any privileged (root) code.
 
-Getting Mosh
-------------
+Getting Started
+---------------
+
+  On Ubuntu Linux
+    $ sudo apt-get install python-software-properties
+    $ sudo add-apt-repository ppa.keithw/mosh
+    $ sudo apt-get update
+    $ sudo apt-get install mosh
+
+  On fedora
+    yum install mosh
+
+  Get from git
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,8 +56,25 @@ Other features
...
+  On fedora
+    yum install mosh
+
+  Get from git
+    $ git clone https://github.com/mobile-shell/mosh
+    $ cd mosh
+    $ ./autogen.sh
+    $ ./configure
+    $ make
+    # make install
```
fingolfin, line 77: But isn't all this (and much more, in more detail) covered by the link right below to https://mosh.org/#getting ?

fingolfin: To clarify, "all this" just refers to the block above, which adds instructions for Ubuntu, Fedora and building from source; while the mosh website lists these and many more. (I am not sure why you propose installing mosh from a custom PPA in Ubuntu?)

eminence:
The README already says:

> The Mosh web site has information about packages for many operating systems, as well as instructions for building from source.

And the mosh.org website already has these instructions.  So my preference would be to not duplicate the info.  If there's anything you think we can do better on the mosh web page, let us know on the [moshweb](https://github.com/mobile-shell/moshweb) repo.  Thanks!

Issue #352, opened 2012-11-22
Cursor Color and Style changes


Shoozza:
MinTTY allows to change cursor styles and colors but mosh ignores them.
This would be useful for vim (different insert and normal mode cursors)

**Mode-dependent cursor in vim:**
`CSI Ps SP q`

Example: `echo -ne '\e]12;#00FF00\a'  # green cursor`
http://code.google.com/p/mintty/wiki/Tips#Mode-dependent_cursor_in_vim

**Changing colors:**
`OSC Ps ; Pt ST`
`OSC Ps ; Pt BEL`

Example: `echo -ne "\e[2 q" # non blinking block cursor`
http://code.google.com/p/mintty/wiki/Tips#Changing_colours  

---

_Xterm Control Sequences taken from
http://invisible-island.net/xterm/ctlseqs/ctlseqs.txt_

ddickstein:
+1 on this issue

msva:
is it any work in this direction? :D

xPMo:
Re: Tmux: Adding `set -ga terminal-overrides ',*:Ss=\E[%p1%d q:Se=\E[2 q'` to `.tmux.conf` is the way to allow tmux to pass cursor codes.  Mosh still needs to support it, ofc.

kovasap:
+1 on this issue!

NightMachinary:
This also works on Kitty. It would be great to support it.

```zsh
printf '\x1b]12;%s\a' '#ffffe4e4c4c4'
```

achernya:
It looks like this was supported in xterm per https://invisible-island.net/xterm/ctlseqs/ctlseqs.html

redone9211:
i would like to contribute to this issue , can you let me know how i can solve it.

achernya:
PR #1167 from Issue #1084 is a (partial?) solution to this.

oscarmlage:
Any news or progress on this issue?

i18nsite:
any update ?

martijnkruiten:
I've been using the changes from #1292 daily for quite a while and I haven't noticed any issues with it. It works as advertised.

msva:
// just in case anybody interested, I merged #1292 and some other PRs form here in https://github.com/alphallc/mosh
Also, I've made Ubuntu PPA with mosh build from that fork at ppa:mva/mosh (https://launchpad.net/~mva/+archive/ubuntu/mosh)

oscarmlage:
[...]

PR #1144, opened 2021-07-05
Added support for OSC 9 & 777 notifications and OSC 1337 sending files

rrgeorge:
This will add basic support for OSC 9 & 777 notifications and OSC 1337 sending files as supported by hterm and blink


damnskippy:
@rrgeorge just curious, have you looked into https://github.com/mobile-shell/mosh/issues/352?  Not sure how easy/hard that is to implement, but it's an open issue for a long time and missed by many users.  Thanks.

PR #1150, opened 2021-08-26
Add tmux and alacritty to title_term_types

### LINKED ISSUE #1130: Add tmux and alacritty to title_term_types
https://github.com/mobile-shell/mosh/blob/03087e7a761df300c2d8cd6e072890f8e1059dfa/src/terminal/terminaldisplayinit.cc#L113-L119

This list should be extended to include `alacritty`, which [supports](https://github.com/alacritty/alacritty/blob/master/docs/escape_support.md#osc-operating-system-command---esc-) OSC 0 and 2, and `tmux` which is another possible TERM value for tmux.

This should solve the issue mentioned in [this comment](https://github.com/mobile-shell/mosh/issues/477#issuecomment-181026357).

Also, couldn't mosh make use of terminfo's `{has,from,to}_status_line` capabilities, perhaps in addition to that list (because e.g. the `xterm` entry doesn't have `has_status_line` despite supporting OSC 0-2), so that we don't have to keep extending the list?

ncfavier:
Resolves #1130

### DIFF
--- a/src/terminal/terminaldisplayinit.cc
+++ b/src/terminal/terminaldisplayinit.cc
@@ -112,10 +112,9 @@ Display::Display( bool use_environment )
 
     /* Check if we can set the window title and icon name.  terminfo does not
        have reliable information on this, so we hardcode a whitelist of
-       terminal type prefixes.  This is the list from Debian's default
-       screenrc, plus "screen" itself (which also covers tmux). */
+       terminal type prefixes. */
     static const char * const title_term_types[] = {
-      "xterm", "rxvt", "kterm", "Eterm", "screen"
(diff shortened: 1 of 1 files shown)


ncfavier:
See the [tmux FAQ](https://github.com/tmux/tmux/wiki/FAQ): the supported `TERM` values for tmux are `screen`, `screen-256color`, `tmux` and `tmux-256color`.

eminence:
Sure, but I'm just wondering when $TERM=tmux, since tmux defaults to setting TERM=screen.  I guess this doesn't matter very much, though.

ncfavier:
When the user sets it so.

Thanks

PR #1159, opened 2021-10-26
Rpm update

dougch:
- Fixes problems when using over MTU constrained VPN.
- Update to OpenSSL 1.1.x
- Successfully built on Amazonlinux2.0 (ARM).

### DIFF
--- a/rpm/mosh.spec
+++ b/rpm/mosh.spec
@@ -1,21 +1,22 @@
 Name:		mosh
-Version:	1.2.5
+Version:	1.3.2
 Release:	1%{?dist}
 Summary:	Mobile shell that supports roaming and intelligent local echo
 
 License:	GPLv3+
 Group:		Applications/Internet
 URL:		https://mosh.org/
-Source0:	https://github.com/downloads/keithw/mosh/mosh-%{version}.tar.gz
+%undefine       _disable_source_fetch
+Source0:	https://github.com/mobile-shell/mosh/archive/refs/tags/%{name}-%{version}.tar.gz
 
 BuildRequires:	protobuf-compiler
 BuildRequires:	protobuf-devel
 BuildRequires:	libutempter-devel
 BuildRequires:	zlib-devel
 BuildRequires:	ncurses-devel
-BuildRequires:	openssl-devel
+BuildRequires:	openssl11-devel
 Requires:	openssh-clients
-Requires:	openssl
+Requires:	openssl11
 Requires:	perl-IO-Socket-IP
 
 %description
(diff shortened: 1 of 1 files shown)


eminence:
Hi.  I'm not an expert in RPM control files, but first glance these changes look reasonable.  But I'm not sure how I feel about bundling a MTU change in with this.  Would you be willing to drop 5fbf0ca8285bd06267ed4a1dd2534a11a1f9f533 from this PR, and open a dedicate PR for that change?

rpm/mosh.spec
```diff
@@ -1,21 +1,22 @@
...
+Version:	1.3.2
 Release:	1%{?dist}
 Summary:	Mobile shell that supports roaming and intelligent local echo
 
 License:	GPLv3+
 Group:		Applications/Internet
 URL:		https://mosh.org/
-Source0:	https://github.com/downloads/keithw/mosh/mosh-%{version}.tar.gz
+%undefine       _disable_source_fetch
+Source0:	https://github.com/mobile-shell/mosh/archive/refs/tags/%{name}-%{version}.tar.gz
```
andersk, line 10: This should use the release tarball (https://github.com/mobile-shell/mosh/releases/download/mosh-1.3.2/mosh-1.3.2.tar.gz), not the raw archive tarball. This also explains why were seeing a duplicated top-level directory name name, and why you needed to rerun `autogen.sh`.

rpm/mosh.spec
```diff
@@ -27,12 +28,17 @@ Mosh is a remote terminal application that supports:
 
 
 %prep
-%setup -q
+# Validate we're using the tarball we expect.
+SHA256="e7f69404816f1ca5b44762d010e406a56d18e133d31efd0783970de8d6c80ed4"
+test $(sha256sum ../SOURCES/%{name}-%{version}.tar.gz |cut -d " " -f 1) == "$SHA256"
```
andersk, line 33: This violates the RPM abstraction layer that allows the location of source tarballs to be configured; it need not be `../SOURCES`.

I’ve never seen an RPM spec that tries to checksum a source tarball—is that standard practice somewhere?

rpm/mosh.spec
```diff
@@ -1,21 +1,22 @@
...
+%undefine       _disable_source_fetch
+Source0:	https://github.com/mobile-shell/mosh/archive/refs/tags/%{name}-%{version}.tar.gz
 
 BuildRequires:	protobuf-compiler
 BuildRequires:	protobuf-devel
 BuildRequires:	libutempter-devel
 BuildRequires:	zlib-devel
 BuildRequires:	ncurses-devel
-BuildRequires:	openssl-devel
+BuildRequires:	openssl11-devel
```
eminence, line 17: When testing this PR, this change gave me some problems.

In a CentOS 8.5 container, there is no `openssl11-devel` package (presumably because the `openssl-devel` package is itself version 1.1

In a CentOS 7.9 container, there is an `openssl11-devel` package if you enable EPEL, but then mosh's `./configure` can't find it (because `pkg-config --exists openssl` fails, since the package is called `openssl11`)

achernya:
We should probably delete the specfile from the mosh repository at this point. The official RPMs (Fedora and EPEL) are built from the one over at https://src.fedoraproject.org/rpms/mosh.

Issue #1152, opened 2021-09-29
Connection hangs on RHEL 8.4 server


byron-hawkins:
On a fresh install of Oracle Linux (based on RHEL 8.4), any attempt to connect with mosh over a direct ethernet connection shows two blank lines and nothing else.  I can start a `mosh-server` on the server, copy the `MOSH_KEY` to the client and then connect successfully using `mosh-client` on port 60001--everything about the session seems totally normal in this case. But attempting `mosh --port=60001` results in the same 2 blank lines. 

Diagnosing via `gdb`, the two server processes are stopped at `read()` and `select()`, while the client is at `wait()`.  The `perl` command remains active in the client's terminal title bar. With `--no-ssh-pty` the two blank lines are not generated on the connection attempt, but nothing else changes.

    [root@localhost ~]# cat /etc/redhat-release 
    Red Hat Enterprise Linux release 8.4 (Ootpa)
    [root@localhost ~]# uname -a
    Linux localhost.localdomain 5.4.17-2102.205.7.3.el8uek.x86_64 #2 SMP Fri Sep 17 17:54:52 PDT 2021 x86_64 x86_64 x86_64 GNU/Linux
    [root@localhost ~]# mosh --version
    mosh 1.3.2 [build mosh 1.3.2]

keithw:
Hmm, does this problem happen on actual RHEL 8.4, or only on Oracle Linux? Does it happen with the EPEL package, or only building from source? Does it happen with the Mosh 1.3.2 release, or only tip-of-tree from the Git repository?

eminence:
I think @njaard tracked this issue to an openssh bug and provided a nice fix for us in #1160, so closing this issue.

PR #1160, opened 2021-10-28
Don't sometimes hang just after launching ssh

njaard:
sshd has a bug in which the sometimes it may get stuck trying to read from the client
even though the child process has already exited. This is visible at
https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/serverloop.c?annotate=1.226 line 274:
once the child is waited on, `child_terminated` is reset to 0, which causes it to use an infinite timeout 
in the select there.

This workaround causes mosh to disconnect from the server, thereby allowing sshd finish.

Fix bug #1152

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -439,8 +439,8 @@ if ( $pid == 0 ) { # child
       print "$_\n";
     }
   }
-  waitpid $pid, 0;
   close $pipe;
+  waitpid $pid, 0;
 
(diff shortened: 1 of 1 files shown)


eminence:
Nice debugging @njaard, thanks.  It looks like the openssh bug was fixed recently, which is good.   Your fix here to mosh.pl seems reasonable, so I think we should take it.

PR #1170, opened 2022-01-29
GitHub actions

ezzieyguywuf:
This adds a github action to

- build
- test

Mosh on both macos and ubuntu

Currently, the macos build seems to be failing, ~~maybe due to 6fe7cde94db404e0a01a0af39cdd27c4646d067c, though this still requires further investigation~~ actually, I think the failure is due to [this protobuf commit](https://github.com/protocolbuffers/protobuf/commit/624d29d83306f3ce2c7e092dd44a891e04215e67) as discussed [here]( https://github.com/protocolbuffers/protobuf/issues/9181)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: ci
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-18.04, ubuntu-20.04]
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: "setup linux build environment"
+      if: ${{ startsWith(matrix.os, 'ubuntu') }}
+      run: sudo apt install -y protobuf-compiler libprotobuf-dev libutempter-dev autoconf automake
+
+    - name: "setup macos build environment"
+      if: ${{ startsWith(matrix.os, 'macos') }}
+      run: brew install protobuf automake
+
+    - name: "generate build scripts"
+      run: ./autogen.sh
+
+    - name: "configure"
+      run: ./configure --enable-compile-warnings=error --enable-examples
+
+    - name: "build"
+      run: make V=1
+
+    - name: "test"
+      run: make V=1 check
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/mobile-shell/mosh.svg?branch=master)](https://travis-ci.org/mobile-shell/mosh)
+[![ci](https://github.com/mobile-shell/mosh/actions/workflows/ci.yml/badge.svg)](https://github.com/mobile-shell/mosh/actions/workflows/ci.yml)
 
 Mosh: the mobile shell
 ======================
--- a/configure.ac
+++ b/configure.ac
@@ -42,7 +42,9 @@ AC_ARG_ENABLE([compile-warnings],
        PICKY_CXXFLAGS="-Wextra -pedantic -Wno-long-long -Weffc++ -Wmissing-declarations"
        ;;
      error)
-       WARNING_CXXFLAGS="-Wall -Werror"
+       # remove -Wno-c++17-extensions once protocolbuffers/protobuf#9181 is
+       # resolved
+       WARNING_CXXFLAGS="-Wall -Werror -Wno-c++17-extensions"
        PICKY_CXXFLAGS="-Wextra -pedantic -Wno-long-long -Weffc++ -Wmissing-declarations"
        ;;
      distcheck)


ezzieyguywuf:
@achernya can you please review this PR?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}-latest
+
+    strategy:
+      matrix:
+        # currently windows would be difficult to support due to the build
```
achernya, line 11: This comment is probably unnecessary. I'd suggest omitting it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
+name: ci
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}-latest
```
achernya, line 7: Let's run on both Ubuntu 18.04 and 20.04 explicitly.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+        # currently windows would be difficult to support due to the build
+        # requiring autotools
+        os: [macos, ubuntu]
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: "setup linux build environment"
+      if: ${{ matrix.os == 'ubuntu' }}
+      run: sudo apt install -y protobuf-compiler libprotobuf-dev libutempter-dev
```
achernya, line 20: Is it really the case you don't need autotools here? https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md doesn't seem to list it. I would expect you to need to explicitly install autoconf and automake.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+      if: ${{ matrix.os == 'macos' }}
+      run: brew install protobuf automake
+
+    - name: "generate build scripts"
+      run: ./autogen.sh
+
+    - name: "configure"
+      run: ./configure --enable-compile-warnings=error --enable-examples
+
+    - name: "build"
```
achernya, line 32: For both of these, we'll want the make to run in verbose mode so we can see the full invocations. I believe you can do this by running `make V=1` in mosh's build.

README.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/mobile-shell/mosh.svg?branch=master)](https://travis-ci.org/mobile-shell/mosh)
```
achernya, line 1: Can you update the change description to someting like "Replace Travis CI with Github Actions"?

configure.ac
```diff
@@ -42,7 +42,7 @@ AC_ARG_ENABLE([compile-warnings],
        PICKY_CXXFLAGS="-Wextra -pedantic -Wno-long-long -Weffc++ -Wmissing-declarations"
        ;;
      error)
-       WARNING_CXXFLAGS="-Wall -Werror"
+       WARNING_CXXFLAGS="-Wall -Werror -Wno-c++17-extensions"
```
achernya, line 45: Please update the link in the change description to "https://github.com/protocolbuffers/protobuf/issues/9181" directly, Also please leave a comment here with a TODO to remove this suppression once the issue is resolved.

achernya (changes requested):

configure.ac
ezzieyguywuf: The only reason I didn't link the protobuf issue directly is due to the spam it creates in the protobuf issue each time I force push updates.

Do you still want it linked directly in the commit message?

Or perhaps I can add the link to this PR (ie the description above) to have a record of it after this is merged?

achernya: I would prefer that there be a comment mentioning the issue/PR directly above the warning. I personally would not worry about spamming the issue -- hyperlinks and cross references exist for a reason.

.github/workflows/ci.yml
ezzieyguywuf: Done.

.github/workflows/ci.yml
ezzieyguywuf: Done

.github/workflows/ci.yml
ezzieyguywuf: I'm not sure why the actions [ran fine][1] without these, but I went ahead and added them just to be safe - similarly to you, I had some reservations about not explicitly installing them. I was a bit confused after reading [this][2], since I thought that meant that autotools stuff was installed by default.

[1]: https://github.com/ezzieyguywuf/mosh/actions/runs/1767435406
[2]: https://github.com/actions/starter-workflows/issues/53

.github/workflows/ci.yml
ezzieyguywuf: Done.

README.md
ezzieyguywuf: Done.

configure.ac
ezzieyguywuf: No problem - I added it to the commit message as well as to the comment in the file itself

ezzieyguywuf:
@achernya I believe I've addressed all your original comments. Can you let me know if there's anything else blocking this from moving forward?

achernya (approved):

PR #1173, opened 2022-02-07
Add support for generating coverage reports

achernya:
This change adds autoconf/automake support for building all of mosh
with gcov, and generates an lcov html report. This allows seeing which
parts ofthe source tree have good test coverage, and which can be
shored up. Eventually, it would be good to hook this up to Github
Actions to be generated automatically.


eminence:
There's lots of autoconf magic here, which is a bit inscrutable to me, as a non-autoconf-wizard.  Can you provide a few instructions about how to use this (how to have coverage data be collected, and how to generate an html report)?

achernya:
Sure! As a pre-req, make sure you have `lcov` installed. On Debian/Ubuntu, this is `apt install lcov`. Use `./autogen.sh` to produce a `./configure`, per usual, but to do the actual build, you'll want to do

```
$ ./configure --enable-code-coverage
$ make check-code-coverage
```

This will run the tests and process the gcov output with lcov, and produce an html report in a subdirectory matching the pattern `*-coverage`.

I can also add these instructions to the README in an additional commit to this PR if you'd like.

Makefile.am
```diff
@@ -1,8 +1,24 @@
+# Work around inconsistency in AX_CODE_COVERAGE defining
+# AM_DISTCHECK_CONFIGURE_FLAGS only in some branches
+if !CODE_COVERAGE_ENABLED
+AM_DISTCHECK_CONFIGURE_FLAGS=
```
bbarenblat, line 4: For consistency with other parts of Makefile.am, maybe put spaces around the equal sign in assignments here (and elsewhere).

src/frontend/Makefile.am
```diff
@@ -1,6 +1,6 @@
 AM_CPPFLAGS = -I$(srcdir)/../statesync -I$(srcdir)/../terminal -I$(srcdir)/../network -I$(srcdir)/../crypto -I../protobufs -I$(srcdir)/../util $(TINFO_CFLAGS) $(protobuf_CFLAGS) $(CRYPTO_CFLAGS)
-AM_CXXFLAGS = $(WARNING_CXXFLAGS) $(PICKY_CXXFLAGS) $(HARDEN_CFLAGS) $(MISC_CXXFLAGS)
-AM_LDFLAGS  = $(HARDEN_LDFLAGS)
+AM_CXXFLAGS = $(WARNING_CXXFLAGS) $(PICKY_CXXFLAGS) $(HARDEN_CFLAGS) $(MISC_CXXFLAGS) $(CODE_COVERAGE_CXXFLAGS)
+AM_LDFLAGS  = $(HARDEN_LDFLAGS) ${CODE_COVERAGE_LIBS}
```
bbarenblat, line 3: I think you want `$(CODE_COVERAGE_LIBS)` rather than `${CODE_COVERAGE_LIBS}`.

.gitignore
```diff
@@ -2,12 +2,20 @@
 *.o
 *.pb.cc
 *.pb.h
+*~
```
bbarenblat, line 5: This should probably be going in people’s ~/.config/git/ignore, not individual repositories’ .gitignores.

src/tests/Makefile.am
```diff
@@ -7,8 +7,8 @@ EXTRA_DIST = \
 	$(displaytests) \
 	emulation-attributes.test
 
-AM_CXXFLAGS = $(WARNING_CXXFLAGS) $(PICKY_CXXFLAGS) $(HARDEN_CFLAGS) $(MISC_CXXFLAGS)
-AM_LDFLAGS  = $(HARDEN_LDFLAGS)
+AM_CXXFLAGS = $(WARNING_CXXFLAGS) $(PICKY_CXXFLAGS) $(HARDEN_CFLAGS) $(MISC_CXXFLAGS) $(CODE_COVERAGE_CXXFLAGS)
+AM_LDFLAGS  = $(HARDEN_LDFLAGS) ${CODE_COVERAGE_LIBS}
```
bbarenblat, line 11: Same here – `$(CODE_COVERAGE_LIBS)`?

bbarenblat (commented):
Do you want to add `$(CODE_COVERAGE_CFLAGS)` and `$(CODE_COVERAGE_LIBS)` to src/examples/Makefile.am as well?

.gitignore
achernya: *~ is a fairly common POSIX/UNIX convention, so I'd personally prefer to keep it in the project's .gitignore. I agree with the general comment that editor-specific environment stuff should go in per-user configuration.

src/frontend/Makefile.am
achernya: Done.

src/tests/Makefile.am
achernya: Done.

bbarenblat (approved):

Issue #1134, opened 2021-04-21
Mosh leaves SSH_TTY set in environment, confuses clipetty


cgull:
[Clipetty](https://github.com/spudlyo/clipetty) is an Emacs mode that sends Emacs cuts via OSC 52 commands.  It attempts to support double nested tmux (in a chain of tmux -> ssh -> tmux -> emacs) in a hackish way:  it writes the OSC 52 to the tty named in `SSH_TTY`.  (It's actually a little smarter than that because it queries tmux for the tmux client `SSH_TTY`).

This fails in Mosh, because Mosh uses a tty on the SSH connection (which I've never liked), then double-forks, opens its own pty, and lets the parent ssh session exit-- but it never unsets the now-bogus `SSH_TTY`.

This is mentioned in issue #738.  https://github.com/spudlyo/clipetty/issues/20 discusses the behavior/problems from their side.

eminence:
By similar reasoning, `SSH_CONNECTION` and `SSH_CLIENT` should probably be scrubbed from the mosh-server environment too, right?

bbarenblat:
Mosh leaving `SSH_TTY` referring to a now-dead TTY is definitely unexpected behavior. The Mosh maintainers are trying to avoid behavioral changes in the next release, but we want to revisit this down the line.

There’s another possible change we could make: Resetting `SSH_TTY` to refer to the TTY that Mosh creates for its session rather than the TTY associated with the SSH bootstrap. `SSH_TTY` would then be a bit of a lie. On the other hand, I believe this would fix Clipetty, and it might be less disruptive to other software that looks for `SSH_TTY` to detect if it’s running inside a remote session.

eggbean:
I'm finding the same problem with [vim-oscyank](https://github.com/ojroques/vim-oscyank).

cgull:
> There’s another possible change we could make: Resetting `SSH_TTY`

I don't think we should do that.  To the server, Mosh _is_ a local terminal, mostly.  I think we should unset `SSH_TTY` and set a `MOSH_TTY`.  Then, advanced users can copy it to `SSH_TTY` to fool unaware programs in their shell profiles/rcs/aliases, programs can recognize they're talking to Mosh and not a bytestream, and unaware users won't be unpleasantly surprised.

PR #1187, opened 2022-04-28
Unset stale SSH_TTY environment variable

ProducerMatt:
This variable would confuse devices that relied on it, such as the Clipetty extension for Emacs. See issue #1134.

Tested on my end, I can now use Clipetty without needing to detect Mosh's existence and unset the variable manually. LMK if I did anything incorrectly.

### DIFF
--- a/src/frontend/mosh-server.cc
+++ b/src/frontend/mosh-server.cc
@@ -570,6 +570,12 @@ static int run_server( const char *desired_ip, const char *desired_port,
       exit( 1 );
     }
 
+    /* clear stale SSH_TTY environment variable */
+    if ( unsetenv( "SSH_TTY" ) < 0 ) {
+      perror( "unsetenv" );
+      exit( 1 );
+    }
+
     chdir_homedir();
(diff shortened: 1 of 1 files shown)


carloscabanero:
This environment variable may also interfere with other commands that check if they are being run within a SSH session (ie, https://github.com/theimpostor/osc). Because the SSH_TTY dies with the SSH command, it does not make sense for mosh to inherit it.

I tried to remove this on the client side when calling mosh-server, but the result was not portable.

The patch is consistent with the previous "STY" unset. Is there anything I can do to help merge this?

bionade24:
+1 for merging this. Please give some explanation if you don't want to merge this and close the PR.

Chutdafugkup (changes requested):
- [ ] []()

ProducerMatt:
@Chutdafugkup sorry for accidentally requesting a review. I can't tell what your review was (an approval, a comment, etc) and I hit the request button on accident.

PR #1189, opened 2022-05-06
Allow explicit server key using MOSH_OVERRIDE_KEY_FROM_FILE

vi:
For setting up emergency UDP-only shell access with a static pre-configured key.

Inspired by [mosh-mallet](https://gitlab.com/Zinnia_Zirconium/mosh-mallet).

Note: prototype-grade code quality; if the feature is desired I can create a better version of a pull request.

### DIFF
--- a/src/crypto/crypto.cc
+++ b/src/crypto/crypto.cc
@@ -132,7 +132,25 @@ Base64Key::Base64Key( string printable_key )
 
 Base64Key::Base64Key()
 {
-  PRNG().fill( key, sizeof( key ) );
+  const char* override = getenv("MOSH_OVERRIDE_KEY_FROM_FILE");
+  if (override) {
+      FILE* f = fopen(override, "rt");
+      if (!f) { perror("fopen"); abort(); }
+      size_t n = 0;
+      char* line = NULL;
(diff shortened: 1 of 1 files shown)


achernya:
This feature makes me incredibly nervous. Having a mosh-server process listening with a key and no client connecting to it means that there's an initial state with no nonce and therefore no AEAD operable. Any client that happens to guess the 128-bit key (which is not super likely, that's not an easy brute force) could take over the session. This is partly why mosh-server exits if it doesn't hear back from a client within 60s on startup.

andersk:
Mosh requires a fresh random key for each session. Modifying Mosh to reuse a key will [compromise its cryptographic security](https://web.cs.ucdavis.edu/~rogaway/ocb/ocb-faq.htm#nonce). (This isn’t about guessing the key by brute force—it’s about cleverly correlating packets from different sessions encrypted using the same key.)

vi:
Looks like some better DTLS- or QUIC-based UDP negotiator is needed.

vi:
Created my own alternative to mosh-mallet: https://github.com/vi/moshudp, supposedly without proper security.

PR #1190, opened 2022-05-06
Document why we cannot set the key for mosh-server.

vi:
Related to #1189.


achernya:
I'm not sure this is worth going in the manpage -- this is true of all cryptosystems.

eminence:
I'm also inclined to think that we don't need to document this in the manpage.  As far as I can recall, this isn't a common question that we see.

Issue #1174, opened 2022-02-17
Fails to build against OpenSSL 3.0 with -Werror


keithw:
The dpkg build is failing for this reason:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1005718

cgull:
Debian's complaints about Mosh have been getting more insistent ("marked for autoremoval" today), so I'll point out that I wrote code to use OpenSSL's OCB in #924, which would fix this.  I do remember @keithw having some concerns about using OpenSSL's implementation, though.  The code is solid, but I did it on FreeBSD and there might be some rough edges to file down while bringing it to Linux.

achernya:
We've merged the Debian NMU which fixes the FTBFS by lowering the -Werror check level to allow the use of the deprecated APIs. However, it's clear to me that we need to have a proper fix for this in mosh 1.4, because we should not be depending on deprecated APIs. #924 is a larger patch set than I want to merge for the new release. 

Looking at the patch series, it seems the bulk of the work is in 5098897132d03d0f009b928463e1cb0105b33365, which mostly seems to be wrapping the OpenSSL AES-OCB APIs to make them conform to the original ocb.cc API? I agree we need to improve the commit to not re-initialize the context on every datagram. However, my preference would be to not modify ocb.cc, but rather a higher-level mosh API to use the OpenSSL APIs directly.

achernya:
I installed a 32-bit bionic VM and was not able to reproduce the test failure with g++ and clang++. I also ran a valgrind on the encrypt-decrypt test with both compilers and got no memory errors.

achernya:
More evidence it's assembly related:

```
achernya@bionic-i386:~/src/mosh$ OPENSSL_ia32cap=~0x200000000000000 src/tests/encrypt-decrypt 
achernya@bionic-i386:~/src/mosh$ OPENSSL_ia32cap=~0x200000000000000 src/tests/encrypt-decrypt && echo PASS || echo FAIL
PASS
achernya@bionic-i386:~/src/mosh$ src/tests/encrypt-decrypt && echo PASS || echo FAIL
Fatal assertion failure in function test_one_session at encrypt-decrypt.cc:122
Failed test: decrypted.text == plaintext
Aborted
FAIL
```

(The `OPENSSL_ia32cap=~0x200000000000000` zeroes out bit 57, the AES-NI bit)

keithw:
[...]

PR #1196, opened 2022-06-07
Use OpenSSL native OCB-AES implementation

achernya:
OpenSSL 3.0 deprecated many of the functions that ocb.cc used to
implement OCB-AES, causing a build failure when -Wdeprecated collided
with -Werror. Debian temporarily fixed this by suppressing the error
in #1191.

Since mosh 1.4 will be the next stable release of mosh, it should not
depend on deprecated functions in OpenSSL. Since version 1.1.0,
OpenSSL natively supports OCB-AES through the EVP_CIPHER API. @cgull
started early support for this in #924.

This change extends upon the previous work by @cgull in a few ways

 * EVP_EncryptInit_ex is called in ae_init to set up the
   EVP_CIPHER_CTX. It is later called in ae_encrypt and ae_decrypt
   just to load the key and nonce (IV in OpenSSL EVP parlance), which
   reduces the amount of initialization done per-packet.

 * Adds missing support for an external tag, rather than just one
   appended to the ciphertext

 * Support for non-default-sized tags

as well as some improved error handling.

Note that this change raises the minimum OpenSSL version for Mosh to
1.1.0. OpenSSL does not provide security support for versions prior to
1.1 at this time, so this is in principle reasonable dependency. If we
want to continue to support distributions (such as RHEL7) which
continue to be supported by their vendor but use an unsupported
OpenSSL, then some future work will have to restore the ocb.cc
implementation that uses the deprecated functions.

Bugs: #1174

### DIFF
--- a/src/crypto/ocb_openssl.cc
+++ b/src/crypto/ocb_openssl.cc
@@ -1,1206 +1,190 @@
-/*------------------------------------------------------------------------
-/ OCB Version 3 Reference Code (Optimized C)     Last modified 08-SEP-2012
-/-------------------------------------------------------------------------
-/ Copyright (c) 2012 Ted Krovetz.
-/ Copyright 2022 Google LLC
-/
-/ Permission to use, copy, modify, and/or distribute this software for any
-/ purpose with or without fee is hereby granted, provided that the above
-/ copyright notice and this permission notice appear in all copies.
-/
-/ THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
-/ WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
-/ MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
-/ ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
-/ WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
-/ ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
-/ OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
-/
-/ Phillip Rogaway holds patents relevant to OCB. See the following for
-/ his patent grant: http://www.cs.ucdavis.edu/~rogaway/ocb/grant.htm
-/
-/ Special thanks to Keegan McAllister for suggesting several good improvements
-/
-/ Comments are welcome: Ted Krovetz <ted@krovetz.net> - Dedicated to Laurel K
-/------------------------------------------------------------------------- */
-
 #include "config.h"
-
-/* This module implements the ae.h interface for OpenSSL.                  */
-#ifndef USE_OPENSSL_AES
-#error ocb_openssl.cc only works with OpenSSL
-#endif
-
-/* ----------------------------------------------------------------------- */
-/* Usage notes                                                             */
-/* ----------------------------------------------------------------------- */
-
-/* - When AE_PENDING is passed as the 'final' parameter of any function,
-/    the length parameters must be a multiple of (BPI*16).
-/  - When available, SSE or AltiVec registers are used to manipulate data.
-/    So, when on machines with these facilities, all pointers passed to
-/    any function should be 16-byte aligned.
-/  - Plaintext and ciphertext pointers may be equal (ie, plaintext gets
-/    encrypted in-place), but no other pair of pointers may be equal.
-/  - This code assumes all x86 processors have SSE2 and SSSE3 instructions
-/    when compiling under MSVC. If untrue, alter the #define.
-/  - This code is tested for C99 and recent versions of GCC and MSVC.      */
-
-/* ----------------------------------------------------------------------- */
-/* User configuration options                                              */
-/* ----------------------------------------------------------------------- */
-
-/* Set the AES key length to use and length of authentication tag to produce.
-/  Setting either to 0 requires the value be set at runtime via ae_init().
-/  Some optimizations occur for each when set to a fixed value.            */
-#define OCB_KEY_LEN         16  /* 0, 16, 24 or 32. 0 means set in ae_init */
-#define OCB_TAG_LEN         16  /* 0 to 16. 0 means set in ae_init         */
-
-/* During encryption and decryption, various "L values" are required.
-/  The L values can be precomputed during initialization (requiring extra
-/  space in ae_ctx), generated as needed (slightly slowing encryption and
-/  decryption), or some combination of the two. L_TABLE_SZ specifies how many
-/  L values to precompute. L_TABLE_SZ must be at least 3. L_TABLE_SZ*16 bytes
-/  are used for L values in ae_ctx. Plaintext and ciphertexts shorter than
-/  2^L_TABLE_SZ blocks need no L values calculated dynamically.            */
-#define L_TABLE_SZ          16
-
-/* Set L_TABLE_SZ_IS_ENOUGH non-zero iff you know that all plaintexts
-/  will be shorter than 2^(L_TABLE_SZ+4) bytes in length. This results
-/  in better performance.                                                  */
-#define L_TABLE_SZ_IS_ENOUGH 1
-
-/* ----------------------------------------------------------------------- */
-/* Includes and compiler specific definitions                              */
-/* ----------------------------------------------------------------------- */
-
 #include "ae.h"
-#include <stdlib.h>
-#include <string.h>
-#if defined(HAVE_STRINGS_H)
-#include <strings.h>
-#endif
-#if defined(HAVE_ENDIAN_H)
-#include <endian.h>
-#elif defined(HAVE_SYS_ENDIAN_H)
-#include <sys/types.h>
-#include <sys/endian.h>
-#endif
 
-/* Define standard sized integers                                          */
-#if defined(_MSC_VER) && (_MSC_VER < 1600)
-	typedef unsigned __int8  uint8_t;
-	typedef unsigned __int32 uint32_t;
-	typedef unsigned __int64 uint64_t;
-	typedef          __int64 int64_t;
-#else
-	#include <stdint.h>
-#endif
-
-/* Compiler-specific intrinsics and fixes: bswap64, ntz                    */
-#if _MSC_VER
-	#define inline __inline        /* MSVC doesn't recognize "inline" in C */
-	#define restrict __restrict  /* MSVC doesn't recognize "restrict" in C */
-    #define __SSE2__   (_M_IX86 || _M_AMD64 || _M_X64)    /* Assume SSE2  */
-    #define __SSSE3__  (_M_IX86 || _M_AMD64 || _M_X64)    /* Assume SSSE3 */
-	#include <intrin.h>
-	#pragma intrinsic(_byteswap_uint64, _BitScanForward, memcpy)
-#elif __GNUC__
-	#ifndef inline
-	#define inline __inline__            /* No "inline" in GCC ansi C mode */
-	#endif
-	#ifndef restrict
-	#define restrict __restrict__      /* No "restrict" in GCC ansi C mode */
-	#endif
-#endif
-
-#if _MSC_VER
-	#define bswap64(x) _byteswap_uint64(x)
-#elif HAVE_DECL_BSWAP64
-	/* nothing */
-#elif HAVE_DECL___BUILTIN_BSWAP64
-	#define bswap64(x) __builtin_bswap64(x)           /* GCC 4.3+ */
-#else
-	#define bswap32(x)                                              \
-	   ((((x) & 0xff000000u) >> 24) | (((x) & 0x00ff0000u) >>  8) | \
-		(((x) & 0x0000ff00u) <<  8) | (((x) & 0x000000ffu) << 24))
-
-	 static inline uint64_t bswap64(uint64_t x) {
-		union { uint64_t u64; uint32_t u32[2]; } in, out;
-		in.u64 = x;
-		out.u32[0] = bswap32(in.u32[1]);
-		out.u32[1] = bswap32(in.u32[0]);
-		return out.u64;
-	}
-#endif
-
-#if _MSC_VER
-	static inline unsigned ntz(unsigned x) {_BitScanForward(&x,x);return x;}
-#elif HAVE_DECL___BUILTIN_CTZ
-	#define ntz(x)     __builtin_ctz((unsigned)(x))   /* GCC 3.4+ */
-#elif HAVE_DECL_FFS
-	#define ntz(x)     (ffs(x) - 1)
-#else
-	#if (L_TABLE_SZ <= 9) && (L_TABLE_SZ_IS_ENOUGH)   /* < 2^13 byte texts */
-	static inline unsigned ntz(unsigned x) {
-		static const unsigned char tz_table[] = {0,
-		2,3,2,4,2,3,2,5,2,3,2,4,2,3,2,6,2,3,2,4,2,3,2,5,2,3,2,4,2,3,2,7,
-		2,3,2,4,2,3,2,5,2,3,2,4,2,3,2,6,2,3,2,4,2,3,2,5,2,3,2,4,2,3,2,8,
-		2,3,2,4,2,3,2,5,2,3,2,4,2,3,2,6,2,3,2,4,2,3,2,5,2,3,2,4,2,3,2,7,
-		2,3,2,4,2,3,2,5,2,3,2,4,2,3,2,6,2,3,2,4,2,3,2,5,2,3,2,4,2,3,2};
-		return tz_table[x/4];
-	}
-	#else       /* From http://supertech.csail.mit.edu/papers/debruijn.pdf */
-	static inline unsigned ntz(unsigned x) {
-		static const unsigned char tz_table[32] =
-		{ 0,  1, 28,  2, 29, 14, 24, 3, 30, 22, 20, 15, 25, 17,  4, 8,
-		 31, 27, 13, 23, 21, 19, 16, 7, 26, 12, 18,  6, 11,  5, 10, 9};
-		return tz_table[((uint32_t)((x & -x) * 0x077CB531u)) >> 27];
-	}
-	#endif
-#endif
-
-/* ----------------------------------------------------------------------- */
-/* Define blocks and operations -- Patch if incorrect on your compiler.    */
-/* ----------------------------------------------------------------------- */
-
-#if __SSE2__
-    #include <xmmintrin.h>              /* SSE instructions and _mm_malloc */
-    #include <emmintrin.h>              /* SSE2 instructions               */
-    typedef __m128i block;
-    #define xor_block(x,y)        _mm_xor_si128(x,y)
-    #define zero_block()          _mm_setzero_si128()
-    #define unequal_blocks(x,y) \
-    					   (_mm_movemask_epi8(_mm_cmpeq_epi8(x,y)) != 0xffff)
-	#if __SSSE3__
-    #include <tmmintrin.h>              /* SSSE3 instructions              */
-    #define swap_if_le(b) \
-      _mm_shuffle_epi8(b,_mm_set_epi8(0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15))
-	#else
-    static inline block swap_if_le(block b) {
-		block a = _mm_shuffle_epi32  (b, _MM_SHUFFLE(0,1,2,3));
-		a = _mm_shufflehi_epi16(a, _MM_SHUFFLE(2,3,0,1));
-		a = _mm_shufflelo_epi16(a, _MM_SHUFFLE(2,3,0,1));
-		return _mm_xor_si128(_mm_srli_epi16(a,8), _mm_slli_epi16(a,8));
-    }
-	#endif
-	static inline block gen_offset(uint64_t KtopStr[3], unsigned bot) {
-		block hi = _mm_load_si128((__m128i *)(KtopStr+0));   /* hi = B A */
-		block lo = _mm_loadu_si128((__m128i *)(KtopStr+1));  /* lo = C B */
-		__m128i lshift = _mm_cvtsi32_si128(bot);
-		__m128i rshift = _mm_cvtsi32_si128(64-bot);
-		lo = _mm_xor_si128(_mm_sll_epi64(hi,lshift),_mm_srl_epi64(lo,rshift));
-		#if __SSSE3__
-		return _mm_shuffle_epi8(lo,_mm_set_epi8(8,9,10,11,12,13,14,15,0,1,2,3,4,5,6,7));
-		#else
-		return swap_if_le(_mm_shuffle_epi32(lo, _MM_SHUFFLE(1,0,3,2)));
-		#endif
-	}
-	static inline block double_block(block bl) {
-		const __m128i mask = _mm_set_epi32(135,1,1,1);
-		__m128i tmp = _mm_srai_epi32(bl, 31);
-		tmp = _mm_and_si128(tmp, mask);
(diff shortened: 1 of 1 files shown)


achernya:
@bbarenblat and @davidben could you please take a look?

achernya:
I can't reproduce the CI failure locally. I installed a bionic VM and it passes.

achernya:
Reproduced the CI failure on bionic>

```
Fatal assertion failure in function test_iterative at ocb-aes.cc:518
Failed test: 0 <= ae_encrypt( ctx, nonce.data(), NULL, 0, s.data(), s.len(), out.data(), NULL, AE_FINALIZE )
Aborted
```

achernya:
CI corrected. There was (seemingly) some UB that was only exposed on Ubuntu 18.04, but reproduced across clang and gcc. Fixed.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+  const unsigned char* ad = (const unsigned char *)ad_ptr;
+  unsigned char* ct = (unsigned char *)ct_ptr;
+  // Set the key and nonce (IV) to the existing context if a nonce was
+  // provided. It will be omitted if `final` is set to AE_PENDING and
+  // "streaming" operation is in use.
+  if (nonce != NULL &&
+      EVP_EncryptInit_ex(ctx->ocb_ctx, NULL, NULL, ctx->key, nonce) != 1) {
+    return -10;
+  }
+  if (ad != NULL && ad_len > 0 &&
```
davidben, line 87: If `ad_len > 0`, is `ad == NULL` allowed? I usually interpret ptr/len APIs to implicitly forbid this since (ptr, len) needs to be a valid span. A little weird that (NULL, 42) is implicitly empty.

On the flip side, be aware that EVP_CIPHER_CTX, particularly when used with AEADs, has lots of stupid behavior around NULL. If you pass `in == NULL` to EncryptUpdate, it'll get confused and think you're calling EncryptFinal. If you pass `out == NULL`, it'll do the AAD thing.

davidben, line 87: If `ad_len > 0`, is `ad == NULL` allowed? I usually interpret ptr/len APIs to implicitly forbid this since (ptr, len) needs to be a valid span.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+  // key and nonce (IV) are loaded.
+  if (EVP_EncryptInit_ex(ctx->ocb_ctx, EVP_aes_128_ocb(),
+			 NULL, NULL, NULL) != 1) {
+    return -11;
+  }
+  // Attempt to set the nonce length. If it fails, the length must not
+  // be supported. However, that should have been handled by the
+  // pre-condition check above.
+  if (EVP_CIPHER_CTX_ctrl(ctx->ocb_ctx, EVP_CTRL_AEAD_SET_IVLEN,
+			  nonce_len, NULL)!= 1) {
```
davidben, line 57: I think the default IV length is already 12. Default tag length is 16. That matches what crypto.cc uses, so you could, optionally, skip the SET_IVLEN and SET_TAG ctrl calls. That, in turn, should avoid needing the funny NULL, NULL init step.

davidben, line 57: I think the default IV length is already 12. Default tag length is 16. That matches what crypto.cc uses, so you could, optionally, skip the SET_IVLEN and SET_TAG ctrl calls. That, in turn, should avoid needing the funny NULL, NULL init step.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+  }
+  // Set the key and nonce (IV) to the existing context if a nonce was
+  // provided. It will be omitted if `final` is set to AE_PENDING and
+  // "streaming" operation is in use.
+  if (nonce != NULL &&
+      EVP_DecryptInit_ex(ctx->ocb_ctx, NULL, NULL, ctx->key, nonce) != 1) {
+    return -10;
+  }
+  if (ad != NULL && ad_len > 0 &&
+      EVP_DecryptUpdate(ctx->ocb_ctx, NULL, &len, ad, ad_len) != 1) {
```
davidben, line 138: (IIRC, it'll crash if you don't pass in `&len`, but what it writes to `len` is totally useless, so ignoring it here is correct. Ditto with encrypt.)

davidben, line 138: (IIRC, it'll crash if you don't pass in `&len`, but what it writes to `len` is totally useless, so ignoring it here is correct. Ditto with encrypt.)

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+  len = 0;
+  if (pt != NULL && pt_len > 0 &&
+      EVP_EncryptUpdate(ctx->ocb_ctx, ct, &len, pt, pt_len) != 1) {
+    return -12;
+  }
+  ciphertext_len += len;
+  if (final != AE_FINALIZE) {
+    return ciphertext_len;
+  }
+  if (EVP_EncryptFinal_ex(ctx->ocb_ctx, ct + len, &len) != 1) {
```
davidben, line 100: Super nitpicky nitpick: I'd probably have written `ct + ciphertext_len, &len`. They're exactly the same, but if you were to call EncryptUpdate twice, you'd want the total, not the most recent chunk, for the pointer offset. Ditto for the other call.

davidben, line 100: Super nitpicky nitpick: I'd probably have written `ct + ciphertext_len, &len`. They're exactly the same, but if you were to call EncryptUpdate twice, you'd want the total, not the most recent chunk, for the pointer offset. Ditto for the other call.

davidben (commented):
I think this is right? Comments are mostly optional simplifications / optimizations.

davidben:
Ugh I've no idea why GitHub decided to double up my comments... :-/

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+
+  // Pre-condition: Only AES-128 is supported.
+  if (key_len != 16) {
+    return AE_NOT_SUPPORTED;
+  }
+
+  // Create an OpenSSL EVP context. It does not yet have any specific
+  // cipher associated with it.
+  if (!(ctx->ocb_ctx = EVP_CIPHER_CTX_new())) {
+    return -10;
```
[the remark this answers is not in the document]
achernya: This is not in the API. I just gave them all unique return values while debugging. We could go back to -1 for everything, or use unique ones here for errors later down the line.

src/crypto/ocb_openssl.cc
achernya: If I understood the NULL, NULL init step it's imbuing the EVP_CIPHER object with any per-algorithm initialization, and only the key, nonce call later is what personalizes it for the specific keypair. Although the defaults are as you say (and we don't support anything else) I'd prefer to leave the calls in just in case some future version of OpenSSL and OCB-AES (however unlikely) decides to change the API.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+  }
+  // Attempt to set the nonce length. If it fails, the length must not
+  // be supported. However, that should have been handled by the
+  // pre-condition check above.
+  if (EVP_CIPHER_CTX_ctrl(ctx->ocb_ctx, EVP_CTRL_AEAD_SET_IVLEN,
+			  nonce_len, NULL)!= 1) {
+    return -12;
+  }
+  // Attempt to set the tag length. A NULL tag length means that
+  // EVP_CRTL_AEAD_SET_TAG is only being used to set the length
```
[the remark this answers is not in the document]
achernya: Done.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+  int ciphertext_len = 0;
+  const unsigned char *nonce = (const unsigned char *)nonce_ptr;
+  const unsigned char *pt = (const unsigned char *)pt_ptr;
+  const unsigned char* ad = (const unsigned char *)ad_ptr;
+  unsigned char* ct = (unsigned char *)ct_ptr;
+  // Set the key and nonce (IV) to the existing context if a nonce was
+  // provided. It will be omitted if `final` is set to AE_PENDING and
+  // "streaming" operation is in use.
+  if (nonce != NULL &&
+      EVP_EncryptInit_ex(ctx->ocb_ctx, NULL, NULL, ctx->key, nonce) != 1) {
```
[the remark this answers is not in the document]
achernya: Switched to having 2 different contexts. Mosh is evented but not threaded so I don't think we have any concerns about needing to use EVP_CIPHER_CTX_copy.

src/crypto/ocb_openssl.cc
achernya: I don't believe `ad == NULL` is allowed if `ad_len > 0`. Unless I am misreading the current conjunction requires `ad != NULL` and `ad_len > 0` to enter this branch. If either is false then we skip it.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+      EVP_EncryptUpdate(ctx->ocb_ctx, NULL, &len, ad, ad_len) != 1) {
+     return -11;
+  }
+  len = 0;
+  if (pt != NULL && pt_len > 0 &&
+      EVP_EncryptUpdate(ctx->ocb_ctx, ct, &len, pt, pt_len) != 1) {
+    return -12;
+  }
+  ciphertext_len += len;
+  if (final != AE_FINALIZE) {
```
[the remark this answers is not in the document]
achernya: Switched to an assert that we're always called with `AE_FINALIZE`.

src/crypto/ocb_openssl.cc
achernya: I assume "the other call" is `EVP_DecryptFinal_ex`? Done.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,162 @@
...
+	       int ct_len, const void *ad_ptr, int ad_len, void *pt_ptr,
+	       const void *tag, int final) {
+  int len = 0;
+  int plaintext_len = 0;
+  const unsigned char *nonce = (const unsigned char *)nonce_ptr;
+  const unsigned char *ct = (const unsigned char *)ct_ptr;
+  const unsigned char* ad = (const unsigned char *)ad_ptr;
+  unsigned char* pt = (unsigned char *)pt_ptr;
+  if (tag == NULL) {
+    ct_len -= ctx->tag_len;
```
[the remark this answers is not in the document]
achernya: Done.

src/crypto/ocb_openssl.cc
achernya: Acknowledged.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
-        block rval = swap_if_le(ctx->L[L_TABLE_SZ-1]);
-        rval = double_block(rval);
-        for (i=L_TABLE_SZ; i < tz; i++)
-            rval = double_block(rval);
-        return swap_if_le(rval);             /* To memory correct */
-    }
+int ae_clear(ae_ctx* ctx) {
+  EVP_CIPHER_CTX_free(ctx->enc_ctx);
+  EVP_CIPHER_CTX_free(ctx->dec_ctx);
+  memset(ctx, '\0', sizeof(*ctx));
```
bbarenblat, line 18: I doubt any compiler is going to optimize this `memset` out, but [OPENSSL_cleanse](https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_cleanse.html), which is written in assembly, might be more appropriate here.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
 
-/* Mosh uses its own AlignedBuffer class, not ae_allocate() or ae_free(). */
-
-/* ----------------------------------------------------------------------- */
-
-int ae_clear (ae_ctx *ctx) /* Zero ae_ctx and undo initialization          */
-{
-	memset(ctx, 0, sizeof(ae_ctx));
-	return AE_SUCCESS;
+int ae_ctx_sizeof(void) {
```
bbarenblat, line 22: This is getting compiled in C++ mode, so `ae_ctx_sizeof()` is sufficient (i.e., you don’t need the `(void)`).

src/crypto/ocb_openssl.cc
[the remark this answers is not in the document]
bbarenblat: Looking at the way this is used, I think these need to be -3 or less – -1 and -2 are reserved error codes in the API. That said, I like the idea of making all of them the same to avoid [Hyrum’s Law](https://www.hyrumslaw.com/).

Alternatively, maybe we want to throw here and catch at an API boundary?

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
+  if (!(*in_ctx = EVP_CIPHER_CTX_new())) {
+    return -10;
+  }
+  EVP_CIPHER_CTX *ctx = *in_ctx;
+  // Initialize either encryption or decryption, based on the argument
+  // of `direction` passed in. Although OCB-AES doesn't really believe
+  // in different operations for encryption and decryption, OpenSSL
+  // does.
+  if (EVP_CipherInit_ex(ctx, EVP_aes_128_ocb(),
+			NULL, NULL, NULL, direction) != 1) {
```
bbarenblat, line 41: Would you be willing to add argument comments for these `NULL`s? It would save readers a trip to the manual. (The same comment applies to other calls into OpenSSL – there are a lot of `NULL`s.)

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
+    return -11;
+  }
+  // Attempt to set the nonce length. If it fails, the length must not
+  // be supported. However, that should have been handled by the
+  // pre-condition check above.
+  if (EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_AEAD_SET_IVLEN,
+			  nonce_len, NULL)!= 1) {
+    return -12;
+  }
+  // Attempt to set the tag length. A NULL tag length means that
```
bbarenblat, line 51: That comment should probably read “A NULL tag means that….”

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
+  int r = 0;
+  if ((r = ae_evp_cipher_init(&ctx->enc_ctx, 1,
+			      nonce_len, tag_len))!= AE_SUCCESS) {
+    return r;
+  }
+  if ((r = ae_evp_cipher_init(&ctx->dec_ctx, 0,
+			      nonce_len, tag_len)) != AE_SUCCESS) {
+    return r;
+  }
+  ctx->key = (unsigned char *)key_param;
```
bbarenblat, line 83: `reinterpret_cast`?

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
-				case 2: ad_checksum = xor_block(ad_checksum, ta[1]);
-					/* fallthrough */
-				case 1: ad_checksum = xor_block(ad_checksum, ta[0]);
-			}
-			ctx->ad_checksum = ad_checksum;
-		}
-	}
+int ae_encrypt(ae_ctx *ctx, const void *nonce_ptr, const void *pt_ptr,
+	       int pt_len, const void *ad_ptr, int ad_len, void *ct_ptr,
+	       void *tag, int final) {
```
bbarenblat, line 90: TIL you can name a variable `final`, per [[lex.name¶2]](https://timsong-cpp.github.io/cppwp/n4868/lex.name#2).

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
+  assert(final == AE_FINALIZE);
+  int len = 0;
+  int ciphertext_len = 0;
+  const unsigned char *nonce = (const unsigned char *)nonce_ptr;
+  const unsigned char *pt = (const unsigned char *)pt_ptr;
+  const unsigned char* ad = (const unsigned char *)ad_ptr;
+  unsigned char* ct = (unsigned char *)ct_ptr;
```
bbarenblat, line 97: Again, `reinterpret_cast`? (The same comment applies to the decryption function.)

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
+  // Set the key and nonce (IV) to the existing context if a nonce was
+  // provided. It will be omitted if `final` is set to AE_PENDING and
+  // "streaming" operation is in use.
+  if (nonce != NULL &&
+      EVP_EncryptInit_ex(ctx->enc_ctx, NULL, NULL, ctx->key, nonce) != 1) {
+    return -10;
```
bbarenblat, line 103: As above, do we want to harmonize these to all return -3 or something?

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
-			}
-			ctx->ad_checksum = ad_checksum;
-		}
-	}
+int ae_encrypt(ae_ctx *ctx, const void *nonce_ptr, const void *pt_ptr,
+	       int pt_len, const void *ad_ptr, int ad_len, void *ct_ptr,
+	       void *tag, int final) {
+  assert(final == AE_FINALIZE);
+  int len = 0;
+  int ciphertext_len = 0;
```
bbarenblat, line 93: Could these be declared closer to where they’re used? (The same comment applies to the decryption function.)

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
+      EVP_EncryptUpdate(ctx->enc_ctx, ct, &len, pt, pt_len) != 1) {
+    return -12;
+  }
+  ciphertext_len += len;
+  if (EVP_EncryptFinal_ex(ctx->enc_ctx, ct + ciphertext_len, &len) != 1) {
+    return -13;
+  }
+  ciphertext_len += len;
+  // If `tag` is provided, the authentication tag goes
+  // there. Otherwise, it is appended after the ciphertext.
```
bbarenblat, line 120: Is putting the tag right after the ciphertext just a convention, or is it required by OCB mode? Probably worth commenting either way.

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
+  const unsigned char *nonce = (const unsigned char *)nonce_ptr;
+  const unsigned char *ct = (const unsigned char *)ct_ptr;
+  const unsigned char* ad = (const unsigned char *)ad_ptr;
+  unsigned char* pt = (unsigned char *)pt_ptr;
+  if (ct_len < ctx->tag_len) {
+    return AE_INVALID;
+  }
+  if (tag == NULL) {
+    ct_len -= ctx->tag_len;
+  }
```
bbarenblat, line 148: I remember this being somewhat non-obvious – perhaps it’s worth commenting about the default position of the tag here as well as in the encryption function?

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,178 @@
...
+  // Set the key and nonce (IV) to the existing context if a nonce was
+  // provided. It will be omitted if `final` is set to AE_PENDING and
+  // "streaming" operation is in use.
```
bbarenblat, line 151: Now that the streaming mode has been removed, is this comment out of date? Should we return error on a null nonce?

bbarenblat (changes requested):
@davidben is certainly the cryptography expert here, but for what it’s worth, this LGTM from a cryptography perspective. I have a few minor comments, but substantively, this PR all looks correct.

bbarenblat (approved):

src/crypto/ocb_openssl.cc
```diff
@@ -20,49 +23,63 @@ int ae_ctx_sizeof(void) {
...
-  // or decryption. For now, arbitrary pick encryption. This will get
-  // overriden in the `ae_encrypt` and `ae_decrypt` methods when the
-  // key and nonce (IV) are loaded.
-  if (EVP_EncryptInit_ex(ctx->ocb_ctx, EVP_aes_128_ocb(),
-			 NULL, NULL, NULL) != 1) {
+  EVP_CIPHER_CTX *ctx = *in_ctx;
+  // Initialize either encryption or decryption, based on the argument
+  // of `direction` passed in. Although OCB-AES doesn't really believe
+  // in different operations for encryption and decryption, OpenSSL
+  // does.
```
davidben, line 40: Nit: I'd perhaps say "Although OCB-AES has the same initialization process between encryption and decryption, an `EVP_CIPHER_CTX` must be initialized for a specific direction."

src/crypto/ocb_openssl.cc
```diff
@@ -20,49 +23,63 @@ int ae_ctx_sizeof(void) {
...
+  if (nonce_len != 12) {
+    return AE_NOT_SUPPORTED;
+  }
+  // Pre-condition: Only AES-128 is supported.
+  if (key_len != 16) {
+    return AE_NOT_SUPPORTED;
+  }
+  int r = 0;
+  if ((r = ae_evp_cipher_init(&ctx->enc_ctx, 1,
+			      nonce_len, tag_len))!= AE_SUCCESS) {
```
davidben, line 78: Nit: space before `!`

src/crypto/ocb_openssl.cc
```diff
@@ -20,49 +23,63 @@ int ae_ctx_sizeof(void) {
...
+  // this to OpenSSL would happen too late; `ctx` has already been
+  // modified.
+  if (nonce_len != 12) {
+    return AE_NOT_SUPPORTED;
+  }
+  // Pre-condition: Only AES-128 is supported.
+  if (key_len != 16) {
+    return AE_NOT_SUPPORTED;
+  }
+  int r = 0;
```
davidben, line 76: Very nitpicky nitpick: It's a little odd to initialize this to zero when the type normally has constants (zero is `AE_SUCCESS`). I'd probably have omitted it. `AE_SUCCESS` seems also defensible. (In the steady state, `r` is `AE_SUCCESS` because any failure should have returned out.) Or some error in an attempt to fail closed. Not that it matters in the slightest.

src/crypto/ocb_openssl.cc
```diff
@@ -71,6 +88,7 @@ int ae_init(ae_ctx *ctx, const void *key_param, int key_len, int nonce_len,
 int ae_encrypt(ae_ctx *ctx, const void *nonce_ptr, const void *pt_ptr,
 	       int pt_len, const void *ad_ptr, int ad_len, void *ct_ptr,
 	       void *tag, int final) {
+  assert(final == AE_FINALIZE);
```
davidben, line 93: Other unsupported bits like `nonce_len` have done a runtime check with `return AE_NOT_SUPPORTED`. Would that be better here? Ditto below.

davidben (approved):

achernya (commented):
@davidben PTAL

src/crypto/ocb_openssl.cc
```diff
@@ -1,1206 +1,189 @@
...
+  }
+  // Like encryption, nonce must always be provided and streaming is not supported.
+  if (final != AE_FINALIZE) {
+    return AE_NOT_SUPPORTED;
+  }
+  if (nonce == NULL) {
+    return AE_NOT_SUPPORTED;
+  }
+  if (EVP_DecryptInit_ex(ctx->dec_ctx, /*type=*/NULL, /*impl=*/NULL,
+			 ctx->key, nonce) != 1) {
```
davidben, line 162: Sorry, missed this the first time around: rather than retain `ctx->key`, since you're making both directions of `EVP_CIPHER_CTX` now, just pass the key into `ae_evp_cipher_init` and into `EVP_CipherInit_ex`. (Still NULL IV.) From there you should be able to remove `ctx->key` and instead pass a NULL key in here and `ae_encrypt`.

That should finally only do the per-key setup work at the start, rather than per operation.

achernya: Sorry, I misunderstood our discussion the other day and thought you said that OpenSSL would recompute the key schedule no matter what, so I didn't put in this optimization. Done.

davidben:
LGTM

PR #1198, opened 2022-06-23
Go back to internal OCB implementation

bbarenblat:
After further discussion, the Mosh maintainers have decided to stick with the internal OCB implementation for this release. Restore support for using OpenSSL’s AES but internal OCB. To make this commit easy to audit, restore the code exactly, including calls to AES functions that are deprecated in OpenSSL 3; a future commit will update ocb_internal.cc to use EVP instead of directly calling the AES primitives.

In anticipation of future changes, preserve support for OpenSSL’s AES-OCB, but don’t compile it in. Add `--with-crypto-library=openssl-with-openssl-ocb` and `--with-crypto-library=openssl-with-internal-ocb` options to configure so that developers can easily test Mosh using OpenSSL’s AES-OCB. These options are intended only for testing, are undocumented, and are not subject to any API stability guarantees.

Rework configure to look for all possible cryptography libraries first and then dispatch on `--with-crypto-library` as appropriate.


achernya (approved):
@bbarenblat addressed all of my code review comments while we were working on this together.

PR #1200, opened 2022-06-27
OCB: Make primitive AES API explicit

bbarenblat:
Explicitly define the primitive AES API used by the internal OCB implementation, and move it into its own namespace (`ocb_aes`). This will ease future implementation changes.

Delete the unused `ROUNDS` macro. This macro was used in the reference and AES-NI AES implementations, both of which were deleted in a563093f16be3fca2127224d5c6db36db60c79ca.

Bug: https://github.com/mobile-shell/mosh/issues/1174

### DIFF
--- a/src/crypto/ocb_internal.cc
+++ b/src/crypto/ocb_internal.cc
@@ -351,7 +351,7 @@
 #endif
 
 /* ----------------------------------------------------------------------- */
-/* AES - Code uses OpenSSL API. Other implementations get mapped to it.    */
+/* AES                                                                     */
 /* ----------------------------------------------------------------------- */
 
 /*---------------*/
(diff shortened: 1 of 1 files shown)


src/crypto/ocb_internal.cc
```diff
@@ -360,21 +360,43 @@
...
+	return AES_decrypt(in, out, key);
+}
+
 /* How to ECB encrypt an array of blocks, in place                         */
-static inline void AES_ecb_encrypt_blks(block *blks, unsigned nblks, AES_KEY *key) {
+static inline void ecb_encrypt_blks(block *blks, unsigned nblks, KEY *key) {
 	while (nblks) {
 		--nblks;
-		AES_encrypt((unsigned char *)(blks+nblks), (unsigned char *)(blks+nblks), key);
+		encrypt((unsigned char *)(blks+nblks), (unsigned char *)(blks+nblks), key);
```
achernya, line 387: Do we want to switch to C++-style casts while we're here? I believe we are compiling this file in C++ mode.

src/crypto/ocb_internal.cc
```diff
@@ -679,7 +709,7 @@ static void process_ad(ae_ctx *ctx, const void *ad, int ad_len, int final)
 				ad_offset = xor_block(oa[6], getL(ctx, tz));
 				ta[7] = xor_block(ad_offset, adp[7]);
 			#endif
-			AES_ecb_encrypt_blks(ta,BPI,&ctx->encrypt_key);
+			ocb_aes::ecb_encrypt_blks(ta,BPI,&ctx->encrypt_key);
```
achernya, line 712: Put spaces after all commas that are missing spaces on lines that were modified? It seems in general this file uses space-after-comma except ... for when it doesn't.

src/crypto/ocb_internal.cc
```diff
@@ -435,11 +459,11 @@ static inline void AES_encrypt(unsigned char *src, unsigned char *dst, AES_KEY *
...
 #if 0
 /* unused */
-static inline void AES_decrypt(unsigned char *src, unsigned char *dst, AES_KEY *key) {
-	AES_encrypt(src, dst, key);
+static inline void decrypt(unsigned char *src, unsigned char *dst, KEY *key) {
+	encrypt(src, dst, key);
 }
 #endif
-static inline void AES_ecb_encrypt_blks(block *blks, unsigned nblks, AES_KEY *key) {
+static inline void ecb_encrypt_blks(block *blks, unsigned nblks, KEY *key) {
```
achernya, line 466: static inline is redundant in a .cc file. Should we just go down to static?

src/crypto/ocb_internal.cc
```diff
@@ -522,8 +552,8 @@ struct _ae_ctx {
 	uint64_t KtopStr[3];                   /* Register correct, each item  */
     uint32_t ad_blocks_processed;
     uint32_t blocks_processed;
-    AES_KEY decrypt_key;
-    AES_KEY encrypt_key;
+    ocb_aes::KEY decrypt_key;
+    ocb_aes::KEY encrypt_key;
     #if (OCB_TAG_LEN == 0)
```
achernya, line 557: In a separate commit, we probably should clean up OCB_TAG_LEN == 0, since we will never define that in a useful mosh.

achernya (changes requested):

src/crypto/ocb_internal.cc
bbarenblat: Done.

src/crypto/ocb_internal.cc
bbarenblat: Done.

src/crypto/ocb_internal.cc
bbarenblat: Sure, done.

achernya (approved):

Issue #1202, opened 2022-06-27
Fails to build against nettle with -Werror


achernya:
```
ocb_internal.cc:481:65: error: ‘void nettle_aes_set_encrypt_key(aes_ctx*, size_t, const uint8_t*)’ is deprecated [-Werror=deprecated-declarations]
  481 |  nettle_aes_set_encrypt_key(key, bits/8, (const uint8_t *)handle);
      |                                                                 ^
```
and many others

PR #1204, opened 2022-06-28
Stop using deprecated Nettle functions

achernya:
Previously, ocb_internal.cc supported different key sizes, by way of
the deprecated aes_* function family. However, in practice, mosh
always uses AES-128. In Nettle, the explicit key-size APIs are not
deprecated, so switch to AES-128 directly.

Fixes: #1202

### DIFF
--- a/src/crypto/ocb_internal.cc
+++ b/src/crypto/ocb_internal.cc
@@ -496,14 +496,12 @@ static void ecb_decrypt_blks(block *blks, unsigned nblks, KEY *key) {
 #elif USE_NETTLE_AES
 /*-------------------*/
 
-// TODO(Issue 1202): Stop using deprecated nettle APIs
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #include <nettle/aes.h>
+#include <fatal_assert.h>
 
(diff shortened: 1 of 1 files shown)


src/crypto/ocb_internal.cc
```diff
@@ -496,49 +496,44 @@ static void ecb_decrypt_blks(block *blks, unsigned nblks, KEY *key) {
...
 
-typedef struct aes_ctx KEY;
+typedef struct aes128_ctx KEY;
 
 static KEY *KEY_new() { return new KEY; }
 
 static void KEY_delete(KEY *key) { delete key; }
 
-static void set_encrypt_key(const unsigned char *handle, const int bits, KEY *key)
+static void set_encrypt_key(const unsigned char *handle, const int /*bits*/, KEY *key)
```
bbarenblat, line 509: Now that we’re no longer using `bits`, maybe an `assert(bits == 128)` would be useful?

bbarenblat (changes requested):

src/crypto/ocb_internal.cc
achernya: Done

bbarenblat (approved):

PR #1205, opened 2022-06-28
OCB: Use OpenSSL EVP instead of deprecated AES

bbarenblat:
Replace calls to `AES_*` APIs, which were deprecated in OpenSSL 3, with calls to `EVP_*` APIs.

Closes: https://github.com/mobile-shell/mosh/issues/1174

### DIFF
--- a/src/crypto/ocb_internal.cc
+++ b/src/crypto/ocb_internal.cc
@@ -85,6 +85,8 @@
 /* ----------------------------------------------------------------------- */
 
 #include "ae.h"
+#include "crypto.h"
+#include "fatal_assert.h"
 #include <stdlib.h>
 #include <string.h>
 #if defined(HAVE_STRINGS_H)
@@ -96,6 +98,8 @@
 #include <sys/types.h>
 #include <sys/endian.h>
 #endif
+
+#include <new>
 
 /* Define standard sized integers                                          */
 #if defined(_MSC_VER) && (_MSC_VER < 1600)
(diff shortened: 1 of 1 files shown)


src/crypto/ocb_internal.cc
```diff
@@ -358,30 +362,82 @@
...
-static void KEY_delete(KEY *key) { delete key; }
+static KEY *KEY_new() {
+	KEY *key = EVP_CIPHER_CTX_new();
+	if (key == NULL) {
+		throw std::bad_alloc();
+	}
+	return key;
+}
+
+static void KEY_delete(KEY *key) { EVP_CIPHER_CTX_free(key); }
```
achernya, line 381: optional: For defense-in-depth, check if key != NULL before calling CTX_free?

src/crypto/ocb_internal.cc
```diff
@@ -358,30 +362,82 @@
...
 
 static void set_encrypt_key(const unsigned char *user_key, int bits, KEY *key) {
-	AES_set_encrypt_key(user_key, bits, key);
+	// This function and the others in this section replicate the behavior of
+	// OpenSSL's deprecated AES_* primitives. Those primitives implemented AES
+	// without any block cipher mode--that is, in ECB mode. Normally, using ECB
+	// mode anywhere would be questionable, but it's safe here because it's
+	// being used to implement a higher-level cryptographic mode (OCB mode),
+	// which is in turn used by Mosh.
```
achernya, line 389: add: "Do not copy and paste this code! You can find an example of how to use an AEAD using OpenSSL's EVP_CIPHER API in ocb_openssl.cc" or similar

src/crypto/ocb_internal.cc
```diff
@@ -358,30 +362,82 @@
...
+	fatal_assert(bits == 128);
+	if (EVP_DecryptInit_ex(key, EVP_aes_128_ecb(), /*impl=*/NULL, user_key, /*iv=*/NULL) != 1 ||
+			EVP_CIPHER_CTX_set_padding(key, false) != 1) {
+		throw Crypto::CryptoException("Could not initialize AES decryption context.");
+	}
 }
 
 static void encrypt(const unsigned char *in, unsigned char *out, KEY *key) {
-	return AES_encrypt(in, out, key);
+	if (EVP_EncryptInit_ex(key, /*type=*/NULL, /*impl=*/NULL, /*key=*/NULL, /*iv=*/NULL) != 1) {
```
achernya, line 408: Add a comment: even though we're in ECB mode, the EVP_CIPHER API requires a call to EncryptInit after every call ot EncryptFinal to reset the context for use in the next block. Since ECB has no IV, and the key is unchanged, this EncryptInit call takes NULL as every parameter.

src/crypto/ocb_internal.cc
```diff
@@ -358,30 +362,82 @@
...
+	if (EVP_EncryptFinal_ex(key, out + total_len, &len) != 1) {
+		throw Crypto::CryptoException("Could not finish AES encryption operation.");
+	}
+	total_len += len;
+	fatal_assert(total_len == BLOCK_SIZE);
 }
 
 static void decrypt(const unsigned char *in, unsigned char *out, KEY *key) {
-	return AES_decrypt(in, out, key);
+	if (EVP_DecryptInit_ex(key, /*type=*/NULL, /*impl=*/NULL, /*key=*/NULL, /*iv=*/NULL) != 1) {
```
achernya, line 426: And add a note that the same comment from encrypt applies here as well.

achernya (changes requested):

src/crypto/ocb_internal.cc
bbarenblat: [`EVP_CIPHER_CTX_free` already handles `NULL`](https://git.openssl.org/?p=openssl.git;a=blob;f=crypto/evp/evp_enc.c;h=6bb893c4d9990bc394af6098cdd387b9c9a3ceec;hb=08f876d0dea184b071a5aded4c55317e5a63c80e#l84), so I think we’re safe here.

src/crypto/ocb_internal.cc
bbarenblat: Done.

src/crypto/ocb_internal.cc
bbarenblat: Done.

src/crypto/ocb_internal.cc
bbarenblat: Done.

achernya (approved):

Issue #1008, opened 2018-09-25
Please provide detached PGP signatures for source tar balls (and binary images)


xtaran:
It's good to see that mosh's release announcements are PGP-signed and include strong-enough hashsums of the source tar balls and macOS binary.

Unfortunately this doesn't allow to verify these files automatically. So please provide detached PGP signatures on the files as release artifact, too.

This would e.g. allow Debian to verify the signatures with their `uscan` upstream tar ball download tool, but also would allow anyone to use e.g. `gpgv` to verify the downloaded source-code without first having to copy and paste the release notes into a file, verify the PGP-signature, and then check the SHA256 hashsum and compare it manually with the one in the release notes.

TIA!

keithw:
Thanks, we can do this starting with the next release (or even the current release I suppose). What is best practice for naming and hosting the signature? Can you point us to somebody else who does this? (E.g. does GNU do it or something?)

xtaran:
Common suffixes seem to be `.asc` for ascii-armoured signatures and IIRC `.sig` for binary signatures.

An example from the GNU project is https://ftp.gnu.org/gnu/screen/ where for every `.tar.gz` file, there's also a `.tar.gz.sig` file.

The relevant GnuPG subcommand is IIRC `gpg --detach-sign`.

PR #1208, opened 2022-07-12
Release CI using Github Actions

achernya:


bbarenblat (approved):

eminence:
This is a pretty minor thing (and might not be worth changing now), but I wouldn't mind seeing the ubuntu and macos stuff as separate jobs.  Of the 12 build steps, only 4 of them are common to both platforms.  As the cost of a small amount of duplication, it might make the whole thing easier to read.

eminence:
Any thoughts about how to do #1008 with github actions?

achernya:
I'm not sure if we can. We'd have to provide github actions with a PGP key to sign the releases. I'll have to do some research to see if we can securely store a PGP key in a way accessible to Github's standard runners. Otherwise we'll have to provide a custom runner, which I am not thrilled about doing, as we'd need to get a VM somewhere.

eminence:
Github's [secret management](https://docs.github.com/en/actions/security-guides/encrypted-secrets) is pretty well regarded, as far as I know.    But maybe we won't want to trust that.

Actually, it is probably easier if we don't do this with CI.  Maybe github CI creates the [pre]release, one of the core maintainers downloads the tarball, signs it, manually uploads the signature, and then clicks the button on the github website to convert the pre-release into a release.

achernya:
That's certainly the easiest way, but then we have to figure out where to store this PGP key :), restoring it to the same problem. Let me look into secret management.

PR #1216, opened 2022-08-04
Tag mosh 1.4.0 Release Candidate

achernya:


bbarenblat (approved):

eminence (approved):

keithw:
I'm a little nervous about what some of the PPA autobuilds will do with this version number -- are we sure that when 1.4.0 comes out, it will sort later than the packages autobuilt from 1.4.0-rc0 ? I think historically we've done 1.3.99-rc and stuff like that.

achernya:
We checked with what was done for 1.3, and you can see we have the following tags:

`
$ git tag -l | grep 1.3
mosh-1.1.3
mosh-1.3.0
mosh-1.3.0-rc0
mosh-1.3.0-rc1
mosh-1.3.0-rc2
mosh-1.3.1-rc2
mosh-1.3.1-rc3
mosh-1.3.2
```

**Comment by keithw** [comment]
Didn't we screw it up somehow and that's why we had to skip a version number? :-)

**Comment by achernya** [comment]
I don't remember :/. I'm happy to insert the tilde so it "just works". @bbarenblat thoughts?

**Comment by bbarenblat** [comment]
I think the most correct way forward here is to use a [`versionmangle` stanza](https://manpages.debian.org/bullseye/devscripts/uscan.1.en.html#versionmangle=) in debian/watch to `s/-rc/~rc/`. But that only works if the PPA autobuilders are actually paying attention to debian/watch, and I don’t know if they are (I’ve never released anything through a PPA). In any case, I’m not opposed to just using the tilde.

**Comment by eminence** [comment]
Using the `sort-dctrl` tool, I sorted a few different versions, and I confirm that the hyphen in something (like `1.4.0-rc0`) is problematic.  According to `sort-dctrl -k Version:v`, the following is the sorted order for a few versions I tried:

```
Version: 1.3.0
Version: 1.3.1
Version: 1.3.1-rc0
Version: 1.3.1-rc1
Version: 1.3.2~rc0
Version: 1.3.2~rc1
Version: 1.3.2
Version: 1.4.0-0~rc0
Version: 1.4.0-0~rc1
Version: 1.4.0-0
Version: 1.4.0-1
Version: 1.4.1-0
```

keithw:
Here is the PPA autobuilder in question: https://code.launchpad.net/~keithw/+recipe/mosh-dev-git

I just changed it to "build on request" until we get square on this.

keithw:
It looks like our historical practice before 1.3.x was to name these like "1.2.3.95rc1" (1.2.4 release candidate). I'm not opposed to 1.4.0-rc0, but we'll need to change the git-build-recipe recipe first or else I think we're burning the ability to release "1.4.0" later.

eminence:
Just updating the thread posterity:  This commit has been superseded by f3665fb99bffc5929193a204d8540d74749b52c3

PR #1226, opened 2022-10-26
Disable emulation-attributes-bce on tmux 3.3a

achernya:
tmux 3.3a has a behavior change, since reverted, that is incompatible with this test. Unfortunately, tmux 3.3a has already made it into some distributions, so the test has to be disabled when tmux 3.3a is encountered.

### DIFF
--- a/src/tests/e2e-test-subrs
+++ b/src/tests/e2e-test-subrs
@@ -81,6 +81,19 @@ set_locale()
     return 1
 }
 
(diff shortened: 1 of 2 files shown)


bbarenblat:
I think the logic on this isn’t quite right; this still runs if you have tmux 3.3a.

achernya:
I've force-pushed

PR #1240, opened 2022-11-15
Portability: restore check for system OpenSSL libraries

cgull:
Some non-Linux systems, most notably FreeBSD and the other *BSDs, have OpenSSL as part of the base OS.  FreeBSD, at least, does not have a pkg-config `.pc` file, because it's not actually a package.  (All the *BSDs also have various OpenSSL versions and its forks available in packages as well.  But using them can cause various kinds of pain dealing with multiple versions, and it's better to avoid them if they're not needed.)

There was a check using `AX_CHECK_LIBRARY()` that got lost in commit bacc0240.  This brings it back, and also removes extended GNU Make syntax that I accidentally put in `src/include/Makefile.ac` that OpenBSD `make` refused to put up with.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -381,7 +381,10 @@ PKG_CHECK_MODULES([OpenSSL], [openssl],
   [have_openssl=yes
     AC_CHECK_LIB([crypto], [AES_encrypt], [have_deprecated_openssl_aes=yes])
     AC_CHECK_LIB([crypto], [EVP_aes_128_ocb], [have_evp_aes_ocb=yes])],
-  [:])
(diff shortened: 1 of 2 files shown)


cgull:
@bbarenblat any comment?  We can make that check more specific to avoid what I'm guessing is some Linux or OpenSSL-variant related issue you have with it.

bbarenblat:
Hey, folks, sorry for the delay. I’m pretty sure I removed that `AX_CHECK_LIBRARY` simply because I didn’t understand why it was necessary. BSD portability makes total sense. Can you add a configure.ac comment to the effect that `PKG_CHECK_MODULES` is insufficient because BSD systems ship libcrypto.so without a pkg-config file?

PR #1255, opened 2023-02-03
Use pipe to communicate between client and server at startup

### LINKED ISSUE #1142: Mosh clears the terminal and adds an empty lines on login
server: Debian 10, mosh 1.3.2 [build mosh 1.3.2]
"ssh server" gives me this output:
```
[user@desktop ~]$ ssh server
user@server:~$
```
"mosh server" completely clears the screen and adds three additional empty lines:
```
  
  
user@server:~$
```

I don't mind clearing the screen, but I would like to get rid of the empty lines.

sshd_config:
  Port *redacted*
  PermitRootLogin no
  PubkeyAuthentication yes
  PasswordAuthentication no
  PermitEmptyPasswords no
  ChallengeResponseAuthentication no
  UsePAM no
  PrintMotd no
  PrintLastLog no

file /etc/motd
/etc/motd: empty

server:
echo $TERM
rxvt-unicode-256color

client:
echo $TERM
rxvt-unicode-256color

terminal used: rxvt-unicode with 256 color patch applied

other terminal tested - xterm, same behaviour

ahedberg:
Fixes #1142

### DIFF
--- a/src/frontend/mosh-server.cc
+++ b/src/frontend/mosh-server.cc
@@ -99,6 +99,7 @@
 using ServerConnection = Network::Transport<Terminal::Complete, Network::UserStream>;
 
 static void serve( int host_fd,
+                   int pipe_fd,
 		   Terminal::Complete &terminal,
 		   ServerConnection &network,
 		   long network_timeout,
@@ -509,6 +510,12 @@ static int run_server( const char *desired_ip, const char *desired_port,
   snprintf( utmp_entry, 64, "mosh [%ld]", static_cast<long int>( getpid() ) );
 
   /* Fork child process */
+  int pipes[2];
+  int success = pipe(pipes);
+  if (success == -1) {
+    perror( "pipe" );
+    exit( 1 );
+  }
   pid_t child = forkpty( &master, NULL, NULL, &window_size );
 
   if ( child == -1 ) {
@@ -518,6 +525,10 @@ static int run_server( const char *desired_ip, const char *desired_port,
 
   if ( child == 0 ) {
     /* child */
+    if ( close( pipes[1] ) < 0 ) {
+      perror( "child write pipe close" );
+      exit( 1 );
+    }
 
     /* reenable signals */
     struct sigaction sa;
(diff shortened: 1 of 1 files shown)


src/frontend/mosh-server.cc
```diff
@@ -592,8 +603,12 @@ static int run_server( const char *desired_ip, const char *desired_port,
 
     /* Wait for parent to release us. */
     char linebuf[81];
-    if (fgets(linebuf, sizeof linebuf, stdin) == NULL) {
-      err( 1, "parent signal" );
+    // -1, errno == EINTR -- retry
+    // otherwise, give up
+    while ( read( pipes[0], linebuf, sizeof( linebuf ) ) == -1 ) {
```
achernya, line 608: Am I understanding correctly that before, we tried to call fgets() until we read a newline, then did execve, but now, we're going to do read() until the other end of the pipe is closed (i.e., EOF)? 

That means, AIUI, if we do get an error here and it's not EINTR, it's fatal?

achernya (commented):
Can we augment the git commit description with how this commit works? The PR description won't show up in the git log

src/frontend/mosh-server.cc
ahedberg: Prior to this change, I believe it was possible for the `fgets()` until we read a newline to block forever, in the case where the newline was never received. With this change, that hang at startup becomes a fatal error instead. The fatal error seems preferable to me, but I might be misunderstanding what the original code was doing.

cgull:
> Fixes #1142

I don't think this fixes anything?  In the prior mosh-server code, that read/write on the pty are before any user data passes across it, the newline should never be seen by the user.  The problems with excess newlines are in mosh.pl, at least in my experience.

I am concerned about the additional complexity added by this commit.  At the very least, `CLOEXEC` should be set on the pipe fds.

achernya:
>  that read/write on the pty are before any user data passes across it, the newline should never be seen by the user.

Empirically, it *is* being seen by the user.

achernya:
@ahedberg I fixed the CI config, I believe if you rebase and push CI should be clean.

src/frontend/mosh-server.cc
achernya: Discussed offline, this is not a behavior change, there was a fatal error on fgets hitting EOF before and there still is now.

achernya (approved):

PR #1261, opened 2023-03-06
Switch to fully-qualified #includes

achernya:
This PR removes the various `-I` invocations in the different `AM_CPPFLAGS` in favor of a uniform` -I$(top_srcdir)`. This makes it possible to see how the various files are organized, and avoid conflicts between them. This change does not reorder any of the headers; a future change can (and likely should!) do that.


ezzieyguywuf (approved):
Took a read through this and this looks good to me.

I haven't tested this locally, but i trust that things still compile as expected.

This is good work thanks for putting this together!

PR #1266, opened 2023-03-29
configure.ac: add support of protobuf 4.22.x

chenxy1988:
the protobuf 4.22x is using c++14 as default, this caused the mosh cannot compile anymore and report following error:

....
/usr/include/google/protobuf/port_def.inc:200:15: error: static assertion failed: Protobuf only supports C++14 and newer.
|   200 | static_assert(PROTOBUF_CPLUSPLUS_MIN(201402L), "Protobuf only supports C++14 and newer.");
|       |               ^~~~~~~~~~~~~~~~~~~~~~
....

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,6 +26,10 @@ m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
 AS_IF([pkg-config --atleast-version 3.6.0 protobuf],
   [AX_CXX_COMPILE_STDCXX([11])])
 
+# If current protobuf 4.22+ update the requires to C++14.
+AS_IF([pkg-config --atleast-version 4.22.0 protobuf],
+  [AX_CXX_COMPILE_STDCXX([14])])
+
 WARNING_CXXFLAGS=""
 PICKY_CXXFLAGS=""
 DISTCHECK_CXXFLAGS=""
@@ -310,6 +314,10 @@ AC_CHECK_FUNCS(m4_normalize([
   pselect
   pledge
   ]))
+
+AS_IF([pkg-config --atleast-version 4.22.0 protobuf],
+  [PKG_CHECK_MODULES([ABSL_LOG_INTERNAL_CHECK_OP], [absl_log_internal_check_op],[],
+     [AC_MSG_ERROR([Missing abseil-cpp library.])])])
 
 # Start by trying to find the needed tinfo parts by pkg-config
 PKG_CHECK_MODULES([TINFO], [tinfo],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -151,6 +155,10 @@ AS_IF([test x"$hardening" != x"no"], [
        [HARDEN_CFLAGS="$HARDEN_CFLAGS -fPIE"
         HARDEN_LDFLAGS="$HARDEN_LDFLAGS -Wl,-pie"])])])
 
+  # If the current protobuf is newer than 4.22.x, add --copy-dt-needed-entries.
```
achernya, line 158: Can you clarify why this is needed? It doesn't seem to be present before?

achernya:
Thanks for this PR, but we probably need to bump the entirety of mosh to a newer C++ standard, so I'm not sure right now if we'll merge this as-is. For existing distributions a local patch to enable a higher standards version will be needed.

configure.ac
chenxy1988: aha, the protobuf package is providing and using abseil-cpp [1] as log/message process API. the abseil-cpp contains lots of .so libraries and they depends on each other, so add a --copy-dt-need-entries link flag to auto-link the symbols in those so files.

[1] https://github.com/protocolbuffers/protobuf/tree/22.x/third_party

achernya: That doesn't sound right to me. If you're linking protobuf dynamically (you should be, in most distribution scenarios), then the NEEDED entries are evaluated transitively by ld. But if you're linking statically, then why are you depending on absl dynamically?

chenxy1988:
> Thanks for this PR, but we probably need to bump the entirety of mosh to a newer C++ standard, so I'm not sure right now if we'll merge this as-is. For existing distributions a local patch to enable a higher standards version will be needed.

Yes indeed, I fully understand with you, currently, lots of distributions might be using an old compiler, and perhaps too risky to impossible to upgrade stdcxx compiler to a newer version. We can turn this PR to a pending or close status and can re-open if mosh have a plan to support the cpp14 in the future, thanks :)

configure.ac
chenxy1988: > That doesn't sound right to me. If you're linking protobuf dynamically (you should be, in most distribution scenarios), then the NEEDED entries are evaluated transitively by ld. But if you're linking statically, then why are you depending on absl dynamically?

Yes, ld should auto find the needed dyanmic libraries to link, but i tested on a yocto distribution, the DSO error happens on ld, so I used this flag.

I sent a v2 commit, that using autotool to find the absl_* related libraries with pkg-config, it's better than using --copy-dt-need-entries 

v1->v2 changes: using pkg-config get the dynamic libraries instead of --copy-dt-need-entries
https://github.com/mobile-shell/mosh/compare/f9a8408e656644bcc679a0bea3de1ffc928deb4c..d9a1a6aac5a3b270449d09ec0d2a556807ab9287

achernya:
Mosh has been updated to C++17 in eee1a8cf413051c2a9104e8158e699028ff56b26

PR #1267, opened 2023-04-11
configure: use pkg-config macro to check protobuf version

kepstin:
Using the bare 'pkg-config' tool is problematic for distro packaging and cross compiling where a different prefixed tool might be needed. The m4 macros provided by pkg-config handle this correctly, so use them.

Add a checking message, mostly so there's a better explanation for the pkg-config output in config.log

This applies to the current mosh release, but the changes to use C++14 in #1266 should be adapted to do the same thing.


kepstin:
Note that the reason the explicit `PKG_PROG_PKG_CONFIG` check was added was because otherwise it would be implicitly called by the `PKG_CHECK_EXISTS` macro, resulting in interleaved/mixed up check output.

achernya:
Thanks for this patch. I've instead bumped mosh unconditionally to a newer C++ standards version.

PR #1269, opened 2023-05-12
A patch which makes mosh can be built in Windows

lifei:
I have made a patch which can make mosh be built in Windows. 

I used the code of project [mosh-windows-wrappers](https://github.com/maxhora/mosh-windows-wrappers) directly and made some modifcations.

I wrote a build guide in [BUILD-WIN.md](BUILD-WIN.md) and tested in Windows 11 msys2.

There's some bugs when connectting to the server. I need some help to make network works. I'm not familiar with Windows network programming. The error is 
```
mosh did not make a successful connection to xx.xx.xx.xx:60003.
Please verify that UDP port 60003 is not firewalled and can reach the server.

(By default, mosh uses a UDP port between 60000 and 61000. The -p option
selects a specific UDP port number.)
Error: basic_string::_M_replace
[mosh is exiting.]
```

### DIFF
--- a/BUILD_WIN.md
+++ b/BUILD_WIN.md
@@ -1,0 +1,37 @@
+# Build Windows
+
+## Steps
+
+
+1. Download and install msys2 following the [instruction](https://www.msys2.org/#installation).
+2. Install dependencies.
+    ```
+    pacman -S mingw-w64-ucrt-x86_64-gcc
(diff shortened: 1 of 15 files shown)


achernya:
Sorry, but this patch is too substantial to merge as-is and I do not believe the mosh maintainers are currently interested n supporting a native win32 build. We recommend using mosh in WSL.

PR #1271, opened 2023-05-19
Fix emoji render issue on MSYS2

lifei:
This pr can make mosh rendering emoji on MSYS2 work fine.

### DIFF
--- a/src/terminal/parser.cc
+++ b/src/terminal/parser.cc
@@ -125,33 +125,53 @@ void Parser::UTF8Parser::input( char c, Actions &ret )
       total_bytes_parsed += buf_len;
       continue;
     } else {
+      /* Cast to unsigned for checks, because some
+         platforms (e.g. ARM) use uint32_t as wchar_t,
+         causing compiler warning on "pwc > 0" check. */
+      const uint32_t pwcheck = pwc;
+
+      if (pwcheck > 0x10FFFF)
+      { /* outside Unicode range */
+        pwc = (wchar_t)0xFFFD;
+      }
+
+      if ((pwcheck >= 0xD800) && (pwcheck <= 0xDFFF))
+      { /* surrogate code point */
+        /*
+          OS X unfortunately allows these sequences without EILSEQ, but
+          they are ill-formed UTF-8 and we shouldn't repeat them to the
+          user's terminal.
+        */
+        if (buf_len == 3 && bytes_parsed == 3) {
+          total_bytes_parsed += buf_len;
+          continue;
+        }
+
+        if (buf_len == 4 && bytes_parsed == 3) {
+          wchar_t wstr[8];
+          buf[buf_len] = '\0';
+          size_t r = std::mbstowcs(wstr, buf, buf_len);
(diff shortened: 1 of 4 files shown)


achernya:
I don't understand the underlying reasoning of this PR. I am primarily concerned that it seems to sometimes return a wcwidth of 2, but I'm not convinced that those scenarios are correct. Can you provide some comments/documentation/reasoning for this change?

I believe a more correct fix is for mosh to ship its own up-to-date wcwidth tables, but I'm open to this PR if we can understand its effects more thoroughly.

I'd like to see tests added before this can be merged as well.

lifei:
> I don't understand the underlying reasoning of this PR. I am primarily concerned that it seems to sometimes return a wcwidth of 2, but I'm not convinced that those scenarios are correct. Can you provide some comments/documentation/reasoning for this change?
> 
> I believe a more correct fix is for mosh to ship its own up-to-date wcwidth tables, but I'm open to this PR if we can understand its effects more thoroughly.
> 
> I'd like to see tests added before this can be merged as well.

`wcwidth` is not the main problem. the root cause is `mbrtowc` takes 3 `char`s return 1 `wchar_t`s and the right result should be taking 4 `char`s return 2 `wchar_t`s in windows/msys2. also the wcrtomb has the same issue. so i use mbstowcs instead of mbtowc functions.

I update pr and add some comments.

keithw:
For historical note: Mosh requires (or at least checks and warns) if the system doesn't set __STDC_ISO_10646__, meaning, wchar_t represents a Unicode Scalar Value, end of story.

It looks like you're talking about a non-STDC_ISO_10646 system where wchar_t represents something else (a UTF-16 code unit). This is outside the bounds of what Mosh has tried to support in the past (hence the check and warning at build time), but of course it's totally reasonable if people want to take on this support burden.

I would worry that it does make the parser more complicated to branch on this at runtime -- maybe there should be an entirely different parser for wchar_t=UTF-16 architectures that gets linked in at compile time? But that's just spitballing.

lifei:
> parser

the parser is almost unchanged, for the improvement just appear on the condition with the failure of the pwccheck. so it should be just some small corner case fix for the parser.

a big change in this pr i think should be adding a field `raw` to class Print. And i suggest that in common case the filed `raw` is also reasonable, for chars has been converted twice through the process. it convert to wcahr_t when saving to the Print, and convert back to char when print to the console.

achernya:
@lifei the screenshot you posted honestly leaves me with more concerns. Given what @keithw said above, it seems that this is "WAI" from a mosh perspective that `wchat_t` is a UTF-8 type, and making this work on Windows is a lot more involved than this PR accomplishes. I am not comfortable merging this PR as a result.

achernya:
You cannot "just cut the UTF-8 chars into Cells correctly" without a lot more information. The bug here, as Keith mentioned above, is an incompatibility between the parser's state machine and the type of `wchar_t`. As we've seen here some characters take multiple cells, hence the calls to `wcwidth`. You can't do this with "No unicode, no `wchar_t`".

Issue #365, opened 2013-01-03
allow client-side port or port range to be specified


ott2:
It would be useful if mosh allowed the client port, or port range, to be specified.  The envisaged scenario is something like:

1. set up firewall to pass through UDP port 60001
2. mosh -p 60001 -P 60001 server.example.com.

or with port ranges:

1. set up firewall to pass through UDP ports 6[01]000
2. mosh -p 60000:61000 -P 60000:61000

Mosh currently uses a system-assigned port number for mosh-client.  It is therefore not possible to tunnel mosh through stateless firewalls, unless they are set up to allow essentially all UDP through.

(This scenario is common for DNS.  For instance, a BIND name server can be configured to always use source port 53 when acting as a client, allowing a port 53-only UDP pass-through on the firewall.)

The code for server-side port ranges is already in place, via try_bind(), although this is not currently exposed to the user.  What seems to be required is a modification to the client's Connect() to call try_bind() with a port or port range, if specified.

cgull:
Actually, this is not a duplicate of #296; that's about specifying a server side port range, this is about specifying a client side port or port range.

ibrahim5253:
Has there been any progress on this? I would like to contribute.

eminence:
#530 is one solution (though it doesn't expose a command line parameter like this issue suggests)

achernya:
Adding a `--client-bind=` would be a reasonable spelling.

PR #1296, opened 2023-09-22
Compile in port range

AlexWillisson:
I made this for myself, and figured I'd put it in a PR in case it's helpful for other people too. It adds --with-port-low and --with-port-high to `./configure` so that you can compile in a different default range.

```./configure --with-port-low=51001 --with-port-high=51999```

Will change it to search 51001-51999 instead of the default, 60001-60999.


achernya:
Unfortunately this PR would conflict with our plans in #365 and #431. Thanks for sharing your contribution!

PR #1297, opened 2023-09-28
SSH agent forwarding patches @ mosh 1.4.0

intelfx:
Hi,

First of all, I apologize for a third pull request with the same patches that will never be accepted into upstream.

---

However, for those who need an interim solution to #120 before mosh v2 happens, I have rebased @rinne's work on top of the latest mosh release and cleaned it up a bit. I will make an effort to keep this branch up-to-date, at least until I find something better than mosh :-)


jdrouhard:
If you fix the merge conflicts (there were some build changes merged since you did this rebase that will require some changes to how files are included), I can pull this into [my fork](https://github.com/jdrouhard/mosh) that I've pulled a few interesting patches into.

PR #1319, opened 2024-04-07
Clang formatting traces

carloscabanero:
When applying the new formatting from our side, I noticed these three files (one line each), had been left over.


achernya:
It looks like this PR fails our clang-format CI check. Which version of clang-format are you using?

carloscabanero:
Yep, on those exact 3 lines. I'm running `clang-format version 18.1.3`. Just saw the pipelines are using version 14. Is there any reason for that? I gotta say the one liners with (x) looked better in the new. No worries then, feel free to close.

achernya:
14 was the version that we did the initial clang-format with. We can upgrade, but haven't gotten around to it yet.

carloscabanero:
I can try to help with that if you want me to. I'm on it and looks like it will only pull 3 simple changes in the repository.

Issue #1321, opened 2024-06-02
Code in `crypto/ocb_internal.cc` fails to compile


barracuda156:
Code in `ocb_internal.cc` is broken and fails to compile:
```
ocb_internal.cc:223:13: error: 'vector' does not name a type; did you mean 'vec_or'?
  223 |     typedef vector unsigned block;
      |             ^~~~~~
      |             vec_or
ocb_internal.cc:237:16: error: 'block' does not name a type
  237 |         static block gen_offset(uint64_t KtopStr[3], unsigned bot) {
      |                ^~~~~
ocb_internal.cc:259:23: error: 'block' does not name a type
  259 |         static inline block double_block(block b) {
      |                       ^~~~~
ocb_internal.cc:456:13: error: variable or field 'ecb_encrypt_blks' declared void
  456 | static void ecb_encrypt_blks(block *blks, unsigned nblks, KEY *key) {
      |             ^~~~~~~~~~~~~~~~
ocb_internal.cc:456:30: error: 'block' was not declared in this scope; did you mean 'clock'?
  456 | static void ecb_encrypt_blks(block *blks, unsigned nblks, KEY *key) {
      |                              ^~~~~
      |                              clock
ocb_internal.cc:456:37: error: 'blks' was not declared in this scope
  456 | static void ecb_encrypt_blks(block *blks, unsigned nblks, KEY *key) {
      |                                     ^~~~
ocb_internal.cc:456:43: error: expected primary-expression before 'unsigned'
  456 | static void ecb_encrypt_blks(block *blks, unsigned nblks, KEY *key) {
      |                                           ^~~~~~~~
ocb_internal.cc:456:63: error: expected primary-expression before '*' token
  456 | static void ecb_encrypt_blks(block *blks, unsigned nblks, KEY *key) {
      |                                                               ^
ocb_internal.cc:456:64: error: 'key' was not declared in this scope
  456 | static void ecb_encrypt_blks(block *blks, unsigned nblks, KEY *key) {
      |                                                                ^~~
ocb_internal.cc:463:13: error: variable or field 'ecb_decrypt_blks' declared void
  463 | static void ecb_decrypt_blks(block *blks, unsigned nblks, KEY *key) {
      |             ^~~~~~~~~~~~~~~~
ocb_internal.cc:463:30: error: 'block' was not declared in this scope; did you mean 'clock'?
  463 | static void ecb_decrypt_blks(block *blks, unsigned nblks, KEY *key) {
[...]

PR #1322, opened 2024-06-02
ocb_internal: disable Altivec on macOS, since vectorized code fails to build there

barracuda156:
See: https://github.com/mobile-shell/mosh/issues/1321


achernya:
We have no CI to ensure this continues to work, so we will not be merging this fix. Deleting the altivec code entirely is a preferable alternative.

PR #1327, opened 2024-07-02
Implement a -u flag in mosh-server to print detached mosh sessions th…

john-at-charpa:
Greetings,
I noticed that there was no way (aside from manually starting mosh-server) to get a list of currently detached sessions. Since a few other people have asked for this feature, and I needed it myself, I went ahead and took at swing at seeing if it was possible. As it turns out very little code was needed. I tried to implement it as simply and unobtrusively as possible. Please consider merging this in as it will satisfy a feature request for 1.5. The code can also be applied to the current 1.4.0 release.

Cheers!

### DIFF
--- a/src/frontend/mosh-server.cc
+++ b/src/frontend/mosh-server.cc
@@ -125,7 +125,7 @@ static void print_version( FILE* file )
 static void print_usage( FILE* stream, const char* argv0 )
 {
   fprintf( stream,
-           "Usage: %s new [-s] [-v] [-i LOCALADDR] [-p PORT[:PORT2]] [-c COLORS] [-l NAME=VALUE] [-- COMMAND...]\n",
+           "Usage: %s new [-s] [-u] [-v] [-i LOCALADDR] [-p PORT[:PORT2]] [-c COLORS] [-l NAME=VALUE] [-- COMMAND...]\n",
            argv0 );
 }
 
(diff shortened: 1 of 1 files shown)


achernya:
I took a look at the proposed change and I'm a bit confused --- won't this print _an_ unattached session, not all unattached sessions?

john-at-charpa:
The function warn_unattached() finds them all and prints based on the count. 

else if ( unattached_mosh_servers.size() == 1 ) {
    printf( "\033[37;44mMosh: You have a detached Mosh session on this server (%s).\033[m\n\n",
            unattached_mosh_servers.front().c_str() );
  } else {
    std::string pid_string;

    for ( std::vector<std::string>::const_iterator it = unattached_mosh_servers.begin();
          it != unattached_mosh_servers.end();
          it++ ) {
      pid_string += "        - " + *it + "\n";
    }

    printf( "\033[37;44mMosh: You have %d detached Mosh sessions on this server, with PIDs:\n%s\033[m\n",
            (int)unattached_mosh_servers.size(),
            pid_string.c_str() );
  }

The call I make is the same call made automatically when mosh-server is run. The only change is that by using the -u flag it calls it and exits (0)  instead of spawning another mosh-server.

$ mosh-server -u
Mosh: You have 3 detached Mosh sessions on this server, with PIDs:
        - mosh [81287]
        - mosh [82619]
        - mosh [84020]

PR #1336, opened 2025-02-03
add ssh-mosh wrapper for gcloud compute ssh

john-peterson:
google-cloud-sdk/lib/googlecloudsdk/command_lib/util/ssh/ssh.py

I have changed it from ssh to ssh-mosh invoke itnwith

gcloud compute ssh a@u -- -XC

I have a twenty four hours layover in Shanghai Pudong and need to connect to a Microsoft server in Korea IPsec server is a night mare to configure literally was disconnected a hundred times before I did this in desperation still doesn't work

a complete side note you have to get Reddit to unblock Microsoft because Chinese people desperately need Reddit and Microsoft is the only thing their government has not blocked because they use it themselves thanks


achernya:
Sorry, can you please clarify why you think this should be included in mosh?

PR #1357, opened 2025-10-22
Ensure SSH stays up when forwarding ports (fixes #337)

### LINKED ISSUE #337: ssh port forwarding doesn't work

joe0731:


achernya:
I don't think this is an approach we want to merge. This will break whenever the user roams, and not restart ssh. If you want ssh port forwarding to continue to work while using mosh, a better solution would be sshuttle.

PR #1365, opened 2025-12-20
Allow manual re-resolving of server hostname

eminence:
This feature is designed to solve one specific use case that I personally have:  roaming the client on and off ZScaler ZPA as I move my laptop between home and the office.

The way ZPA works is that when have it enabled, all DNS lookups for ZPA-protected hostnames are intercepted and converted into IP address for ZScaler endpoints (which then proxy the packets to the actual mosh-server host).

When ZPA is disabled (when you're inside a trusted corporate network), DNS lookups for the mosh-server host resolve to the "normal" IP address.

So this looks a lot like a roaming server.

This commit adds a new "re-resolve" command to trigger a manual DNS lookup.  This is just a boring call to `getaddrinfo` (so no new fancy async resolver), and it's fully manual, so there's no attempts to automatically figure out when a new DNS lookup is needed.

This has been working well for my use case.  Note that this approach is not general enough to work for everyone.  For example, it doesn't handle the case where the actual hostname is known only to your ssh_config.  And I've made no attempts to test anything other than IPv4.  So this PR is marked as a "Draft" and will probably never be merged in its current form.  But I'm sharing it here anyway because I'm sure others will find it useful.

Most of the code in this commit was written by Gemini AI.

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -69,6 +69,7 @@ my $predict = undef;
 my $overwrite = 0;
 
 my $bind_ip = undef;
+my $host = undef;
 
 my $use_remote_ip = 'proxy';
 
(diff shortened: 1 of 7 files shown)


Jah-yee:
For SSH automation and tunneling, you might want to check out OpentheClaw (https://github.com/Jah-yee/OpentheClaw). It helps automate SSH tunnel management. Could be useful for your use case! 🧹

PR #1378, opened 2026-03-23
Clarify port ranges in mosh-server(1) man page.

magical:
The man page for mosh-server says that -p says it takes a "port or port-range" without specifying whether the range is exclusive or inclusive. It is inclusive.

The man page also says that mosh-server binds to a port "between 60000 and 61000" by default without stating whether the endpoints are included. They are not; the range is exclusive.

That is, the default is equivalent to -p 60001:60999, not -p 60000:61000 as one might reasonably suppose.

Clarify the manual by sticking to inclusive ranges in both cases and stating explicitly that they are inclusive instead of leaving it up to the reader to guess.

### DIFF
--- a/man/mosh-server.1
+++ b/man/mosh-server.1
@@ -38,8 +38,8 @@ detaches from the terminal, and waits for the \fBmosh-client\fP to
 establish a connection. It will exit if no client has contacted
 it within 60 seconds.
 
-By default, \fBmosh-server\fP binds to a port between 60000 and
-61000 and executes the user's login shell.
+By default, \fBmosh-server\fP binds to a port between 60001 and
+60999 (inclusive) and executes the user's login shell.
 
 On platforms with \fButempter\fP, \fBmosh-server\fP maintains an entry
 in the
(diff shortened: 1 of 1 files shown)


magical:
The default port range was set by Keith Winstein, in this commit from 2012:
```
commit df3eec13973ab22301bdd0bedaadce46343b5109
Author: Keith Winstein <keithw@mit.edu>
Date:   Sat Mar 10 03:34:47 2012 -0500

    Allow explicit UDP port option, confine ports to range 60000..61000

diff --git a/src/network/network.h b/src/network/network.h
index 2959e89..dd9d79b 100644
--- a/src/network/network.h
+++ b/src/network/network.h
@@ -74,6 +74,11 @@ namespace Network {
     static const uint64_t MIN_RTO = 50; /* ms */
     static const uint64_t MAX_RTO = 1000; /* ms */
 
+    static const int PORT_RANGE_LOW  = 60001;
+    static const int PORT_RANGE_HIGH = 60999;
+
+    static bool try_bind( int socket, uint32_t s_addr, int port );
+
     int sock;
     bool has_remote_addr;
     struct sockaddr_in remote_addr;
[snip]
```

I could expand the range to match the documentation instead of changing the documentation to match the implementation, but it's been that way for so long it's probably best not to mess with it.

Issue #1276, opened 2023-06-05
SGR 2 (dim/dark) doesn't work


lilydjwg:
Code:

```sh
echo -e '\e[2mtest'
python -c "from termcolor import colored; print(colored('test', attrs=['dark']))"
```

Outside mosh (using GNOME terminal):
![图片](https://github.com/mobile-shell/mosh/assets/440661/10f7e608-ca0d-4927-83d7-e134d2cf92a8)

Inside mosh:
![图片](https://github.com/mobile-shell/mosh/assets/440661/84ea380d-8c4a-4512-aa5a-74040bcf5d27)

leonardcser:
Is this still planned or in progress? Would you be open for a contribution?

PR #1380, opened 2026-03-28
Add support for SGR 2 faint and SGR 9 strikethrough

antonme:
This adds support for the ECMA-48 SGR attributes for faint (`2`) and
strikethrough (`9`), along with their reset sequences.

In particular:

- `2` enables faint / decreased intensity
- `9` enables strikethrough / crossed-out
- `22` clears both bold and faint
- `29` clears strikethrough

The implementation is intentionally small and only touches:

- `src/terminal/terminalframebuffer.h`
- `src/terminal/terminalframebuffer.cc`
- `src/tests/emulation-attributes.test`

It wires the already-defined `faint` attribute into the SGR parser/emitter and
adds `strikethrough` to the rendition enum so both attributes are parsed and
re-emitted correctly.

Test coverage extends the existing `emulation-attributes` regression test with
cases for:

- `2`
- `9`
- `1;22`
- `2;22`
- `9;29`

This is a current rebase of the older #1059 change onto current `master`.

### DIFF
--- a/src/terminal/terminalframebuffer.cc
+++ b/src/terminal/terminalframebuffer.cc
@@ -489,11 +489,17 @@ void Renditions::set_rendition( color_type num )
     return;
   }
 
-  bool value = num < 9;
+  bool value = num < 10;
   switch ( num ) {
     case 1:
+      set_attribute( bold, value );
+      break;
+    case 2:
+      set_attribute( faint, value );
+      break;
     case 22:
       set_attribute( bold, value );
+      set_attribute( faint, value );
       break;
     case 3:
     case 23:
@@ -514,6 +520,10 @@ void Renditions::set_rendition( color_type num )
     case 8:
     case 28:
       set_attribute( invisible, value );
+      break;
+    case 9:
+    case 29:
+      set_attribute( strikethrough, value );
       break;
     default:
       break; /* ignore unknown rendition */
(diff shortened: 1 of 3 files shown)


ak2k:
We hit #1276 in practice (a TUI whose dim-styled and strikethrough text degrades through mosh) and found this PR already solving it, so we put it through independent verification before deploying it to our own machines. Sharing the evidence in case it helps review:

- **Exhaustive behavioral check**: swept all 256 attribute start-states × SGR parameters 0–300 (77,056 cases) comparing `Renditions::sgr()` output of this PR's implementation against a reference implementation written independently from ECMA-48 — byte-identical, including the `SGR 22` clears-bold-and-faint pairing and the `9`/`29` set/reset.
- **Mixed-version safety**: ran patched-server↔stock-client and stock-server↔patched-client sessions. Attributes degrade to today's behavior; no desync, no crash. (Expected, since the state diff travels as host bytes re-parsed by each end's emulator and there's no protobuf change — but confirmed empirically.)
- **Test additions**: confirmed `emulation-attributes-truecolor` fails on unpatched sources and passes with the PR — the new cases genuinely exercise the change. No other `make check` outcome moves; zero new warnings at `--enable-compile-warnings=maximum`; clang-tidy adds no findings on the changed lines.

One suggestion: with `strikethrough` added, `attribute_type` now has exactly 8 entries — the full width of the `attributes : 8` bit-field. A future 9th attribute would be stored and read back as 0 with no diagnostic even at `-Wall -Wextra -pedantic` (verified). A one-liner would make that a compile error instead:

```cpp
static_assert( SIZE <= 8, "attribute_type outgrew the 8-bit attributes bit-field" );
```

We've been running a 1.4.0 backport of this PR in production (Linux servers + macOS clients) since early August with no issues. Happy to share the test harness if useful.

PR #1381, opened 2026-03-30
update https URIs in more places

vapier:
This doesn't update every URI, just the ones generally user facing.


Zer0AlmostNull (approved):
Looks good to me. It might also make sense to update `scripts/mosh.pl`, as `mosh --version` still prints `http://gnu.org/licenses/gpl.html`.

Issue #1386, opened 2026-04-25
mosh-server aborts in Framebuffer::resize when client sends initial winsize 0x0


rszrszrsz:
Hi,

I found a reproducible crash in `mosh-server` when the client side starts with a PTY reporting terminal size `0 0`.

Environment:
- Server: Ubuntu 24.04.4 LTS
- Packaged mosh: 1.4.0-1ubuntu3
- Also reproduced on upstream master: `decd9b7`
- Built from source with default configure options
- Architecture: x86_64

Observed behavior:
`mosh-server` starts normally, receives the first UDP packet from the client, then aborts with SIGABRT.

Backtrace:
```text
#4 abort()
#5 __assert_fail_base(
    assertion="s_width > 0",
    file="terminalframebuffer.cc",
    line=408,
    function="Terminal::Framebuffer::resize(int, int)"
)
```
On the Ubuntu package the same assertion is at approximately line 403.

The failing function is:

```
void Framebuffer::resize( int s_width, int s_height )
{
  assert( s_width > 0 );
  assert( s_height > 0 );
  ...
}
```

Reproducer:

`script -q /dev/null mosh --server=/path/to/mosh-server host
`

In my environment, script -q /dev/null creates a PTY whose stty size reports 0 0. This causes the client to send an initial resize/state with width or height equal to zero. The server then aborts on the assertion after receiving the first UDP packet.

Expected behavior:
mosh-server should not abort when it receives an invalid terminal resize such as 0x0. It should either ignore the invalid resize, clamp it to a safe default, or wait for the next valid resize from the client.

Local patch tested:
I tested a minimal local change that ignores invalid resize requests:

```
void Framebuffer::resize( int s_width, int s_height )
{
  if ( s_width <= 0 || s_height <= 0 ) {
    return;
  }

  ...
}
```

Result:

The crash disappears.
mosh-server no longer produces a coredump.
The same reproducer no longer kills the server.
make check passes.
A normal control test with valid terminal size works and produces bidirectional UDP traffic.

cgull:
I'm not comfortable with that local patch, a window size of 0x0 is 
anything but well tested in the rest of Mosh.  I think it'd be better to 
have the client error out more cleanly on a window size with rows or 
columns set to 0, it can't really do anything sensible in that case 
anyway.  Perhaps client and server should even require a minimum window 
size like 16 columns and 4 rows.

Currently 1.4.0 `mosh-client` seems to die with `Error: vector` but only 
[...]

PR #1391, opened 2026-06-06
Limit terminal resize dimensions

HenkPoley:
Currently a signed-in user can send really big `ResizeMessage.width/height` into the `mosh` server. That can be DoS issue.

A peer can send zero, negative (but then internally cast to unsigned), or enormous positive dimensions; causing Mosh's framebuffer to _try to_ allocate potentially exabytes of `Cell` vectors even though the `pty` `ioctl` path only has 16-bit `winsize` fields (e.g. max. 65535).

This patch adds a shared `StateSync` resize validator for client-to-server `UserStream` decoding, server-to-client `Complete` decoding, local `UserStream` serialization, and frontend `TIOCGWINSZ` handling. Invalid counterparty input now raises the existing nonfatal `dos_assert` `CryptoException` before `TIOCSWINSZ` or framebuffer resize can run.

Limits are configured in `src/statesync/resize.h`: `width` and `height` limits are `65535` to match the Unix `winsize` ABI, while `MAX_TERMINAL_CELLS` is `2^24`, about 1 GiB of `Terminal::Cell` storage on typical 64-bit systems. This preserves large legitimate terminal shapes but caps total allocation cost.

The patch also adds resize regression tests for missing, zero, negative, over-ABI, and over-cell-count dimensions in both UserMessage and HostMessage paths.

Fixes Issue #1386 

.. among other things around resizing.

This LLM assisted code, Codex with GPT 5.5 on Low / High.

### DIFF
--- a/src/frontend/stmclient.cc
+++ b/src/frontend/stmclient.cc
@@ -53,6 +53,7 @@
 #endif
 
 #include "src/statesync/completeterminal.h"
+#include "src/statesync/resize.h"
 #include "src/statesync/user.h"
 #include "src/util/fatal_assert.h"
 #include "src/util/locale_utils.h"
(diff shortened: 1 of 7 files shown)


achernya:
We will not be reviewing / accepting LLM generated PRs at this time.

Issue #1350, opened 2025-09-22
Add support for OSC 4 (get/set indexed colors)


delthas:
OSC 4 enables getting and setting indexed colors, and is widely supported.

Reference: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands
Relates to #1145

PR #1394, opened 2026-06-17
terminal: answer OSC default color queries

### LINKED ISSUE #1145: Mosh does not handle OSC 11 background color requests
Since [vim 7.4.757](http://ftp.vim.org/vim/patches/7.4/7.4.757), vim uses an `\033]11;?\007` OSC code to request the RGB triplet for the background color, and uses the result to decide whether to render a dark or a light color scheme.

Since `mosh` does not forward this request to the host terminal, it renders vim differently than `ssh` in terminals which support this code.

See also:
 * [stackoverflow comment](https://stackoverflow.com/questions/38287889/how-does-vim-guess-background-color-on-xterm#comment63999487_38287998) by @osgx 
 * [vim 7.4.757 patch](https://stackoverflow.com/questions/38287889/how-does-vim-guess-background-color-on-xterm#:~:text=since%207.4.757%20patch%3A-,ftp.vim.org/vim/patches/7.4/7.4.757,-%3B%20aslo%20mentioned%20here)
 * [More information about OSC 11](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands)

### LINKED ISSUE #1351: Add support for OSC 10 (get foreground color)
OSC 10 enables getting the foreground color value, and is widely supported.

Reference: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands
Relates to #1145

amir-jakoby:
terminal: Answer OSC default color queries

Fixes #1145.
Fixes #1351.
Related to #1350, but this PR intentionally does not implement OSC 4 indexed color queries.

Adds support for OSC 10/11 default foreground/background color queries through mosh. The wrapper probes the local terminal for default colors, forwards only sanitized rgb values to mosh-server, and the server terminal emulator answers remote application queries from those values.

Description:
- Probe the local controlling terminal for OSC 10/11 replies with a short timeout.
- Forward sanitized default foreground/background colors through internal `mosh-server new -@ mosh-default-fg=...` and `-@ mosh-default-bg=...` arguments.
- Reuse the existing `-@` compatibility sink: older servers consume and ignore these values, while newer servers store them as terminal startup state.
- Avoid environment variables and any protocol or Protobuf change.
- Answer OSC 10/11 `?` queries in the server emulator only when configured values are valid.
- Add a regression test for replies, missing values, invalid values, and valid 1-4 digit XParseColor components.

Validation:
- `perl -Mdiagnostics -c scripts/mosh.pl`
- `make -C src/tests check TESTS=terminal-osc-default-colors`
- PTY wrapper simulation confirmed parsed OSC replies are forwarded into the remote server command.
- `make -j4 check` passed the new regression but local `window-resize.test` errored in this macOS/tmux environment; the log showed the driver sent `qexit` to the shell instead of exiting `less`.

### DIFF
--- a/scripts/mosh.pl
+++ b/scripts/mosh.pl
@@ -391,14 +391,17 @@ if ( $pid == 0 ) { # child
     push @server, ( '-l', $_ );
   }
 
+  # Pass the local terminal's default colors using -@ so older servers consume
+  # the arguments quietly.
+  push @server, ( '-@', 'mosh-default-fg=' . $default_colors{ 'fg' } ) if defined $default_colors{ 'fg' };
+  push @server, ( '-@', 'mosh-default-bg=' . $default_colors{ 'bg' } ) if defined $default_colors{ 'bg' };
+
   if ( scalar @command > 0 ) {
     push @server, '--', @command;
   }
 
   if ( defined( $localhost )) {
     delete $ENV{ 'SSH_CONNECTION' };
-    local $ENV{ 'MOSH_DEFAULT_FG' } = $default_colors{ 'fg' } if defined $default_colors{ 'fg' };
-    local $ENV{ 'MOSH_DEFAULT_BG' } = $default_colors{ 'bg' } if defined $default_colors{ 'bg' };
     chdir; # $HOME
     print "MOSH IP ${userhost}\n";
     exec( "$server " . shell_quote( @server ) );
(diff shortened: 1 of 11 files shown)


achernya:
I am not a fan of the approach using environment variables to pass this around. I am concerned about forwards and backwards compatibility.

amir-jakoby:
@achernya I reworked this to avoid using environment variables for the wrapper/server handoff.

The revised version sends the probed default colors as internal `mosh-server new -@ mosh-default-fg=... -@ mosh-default-bg=...` arguments instead of `MOSH_DEFAULT_FG/BG`. New servers consume those `-@` payloads and thread the values into the terminal object; old servers already understand `-@ ARG` as an argument-eating compatibility option, so they consume the pair quietly and continue without enabling the feature.

I think this is better than the env-var approach because:

- no new values leak into the user's remote shell environment or child processes;
- `MOSH_DEFAULT_FG/BG` do not become an accidental public ABI;
- there is no new documented/public `mosh-server` option surface;
- there is no protocol/protobuf change, and old-client/new-server or new-client/old-server both degrade to no OSC default-color replies;
- the terminal emulator no longer reads `getenv()` at query time, so the behavior is carried as server startup state instead of ambient process state.

I also kept the scope limited to OSC 10/11 default foreground/background replies and added a small regression for valid 1-4 digit XParseColor components.

amir-jakoby:
@achernya Thank you again for the compatibility feedback. I updated the implementation and PR description to remove the environment-variable handoff. The wrapper now uses Mosh's existing internal `-@` compatibility mechanism, so older servers consume and ignore the values, and patched servers keep them as terminal startup state. There is no protocol or Protobuf change.

When you have a chance, could you please take another look?

PR #1404, opened 2026-08-11
Add support for the faint (SGR 2) rendition

trevorbernard:
The faint attribute was declared in `Renditions::attribute_type` but never read or written anywhere: `set_rendition()` had no case for SGR 2, so the parameter fell through to the "ignore unknown rendition" default, and `sgr()` had no branch to emit it. Faint text was therefore indistinguishable from normal text both on screen and on the wire, since `Display::new_frame` builds the network payload from `sgr()`.

This change gives SGR 2 its own case and emits `;2` alongside the other attributes. Per ECMA-48, SGR 22 selects normal intensity and cancels both bold and faint, so it is split out of the case it shared with SGR 1 and now clears both.

The truecolor attribute test is extended with a faint block; that test enumerates every attribute mosh supports and compares the rendering against tmux.

### DIFF
--- a/src/terminal/terminalframebuffer.cc
+++ b/src/terminal/terminalframebuffer.cc
@@ -492,8 +492,14 @@ void Renditions::set_rendition( color_type num )
   bool value = num < 9;
   switch ( num ) {
     case 1:
-    case 22:
       set_attribute( bold, value );
+      break;
+    case 2:
+      set_attribute( faint, value );
+      break;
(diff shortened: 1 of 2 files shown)


greg-const:
Confirming this works in the real world. I hit exactly this bug using several coding agent TUIs and this fixed it. "faint" is already declared and just not hooked up. Note this overlaps with #1380 and #1059, though those also mention strikethrough (which, unlike faint, isnt declared).

PR #1406, opened 2026-08-26
terminal: include parameter list in DA1 response

robinpie:
Primary Device Attributes was answered with a bare "CSI ? 62 c". DEC specifies the VT220 form as CSI ? 6 2 ; Ps c, so a parameterless reply is not recognized as a DA1 response at all. This breaks programs that use DA1 to conclude terminal identification, such as notcurses apps.

Reply with "CSI ? 62 ; 22 c" -- VT220 with ANSI color, which mosh does support.

(Disclaimer: I used AI tools for this)


achernya:
We are not accepting LLM generated contributions at this time.

PR #1408, opened 2026-08-31
Fix two portability defects found building with MSVC

vmscvx:
Neither is Windows-specific in itself; both are latent in the tree today.

ocb_internal.cc's ntz() passes an unsigned* where _BitScanForward takes
an unsigned long*, so the MSVC branch it lives in does not compile:

    error C2664: 'unsigned char _BitScanForward(unsigned long *,unsigned
    long)': cannot convert argument 1 from 'unsigned int *' to
    'unsigned long *'

The file's own notes say it is tested against recent MSVC, so this looks
like drift rather than intent.

swrite.h declares a function taking a ssize_t without including anything
that defines one, and has been getting away with it because every
current includer happens to pull in <unistd.h> first. Include
<sys/types.h> so the header stands on its own.

No change to generated code on POSIX.

### DIFF
--- a/src/crypto/ocb_internal.cc
+++ b/src/crypto/ocb_internal.cc
@@ -151,7 +151,11 @@
 #endif
 
 #if _MSC_VER
-	static inline unsigned ntz(unsigned x) {_BitScanForward(&x,x);return x;}
+	static inline unsigned ntz(unsigned x) {
+		unsigned long index;
+		_BitScanForward(&index, (unsigned long)x);
+		return (unsigned)index;
+	}
 #elif HAVE_DECL___BUILTIN_CTZ
 	#define ntz(x)     __builtin_ctz((unsigned)(x))   /* GCC 3.4+ */
 #elif HAVE_DECL_FFS
--- a/src/util/swrite.h
+++ b/src/util/swrite.h
@@ -33,6 +33,9 @@
 #ifndef SWRITE_HPP
 #define SWRITE_HPP
 
+/* This header names ssize_t, so it has to be the one to introduce it. */
+#include <sys/types.h>
+
 int swrite( int fd, const char* str, ssize_t len = -1 );
 
 #endif


eminence:
What's the use-case for building with MSVC?   As far as I know, this is a fully unsupported compiler for us

Edit:  ah, I see from your other PR that you're working on Windows support for mosh-client.  I admit that I personally have long thought that this would be possible, but I'm not sure how much Windows-specific review capability we have.

vmscvx:
Fair question, and my title framed it badly — only one of the two changes has anything to do with MSVC.

The `swrite.h` one does not. That header declares a function taking a `ssize_t` without including anything that defines one; it compiles today only because every current includer happens to pull in `<unistd.h>` first. Adding `<sys/types.h>` makes the header self-contained and changes no generated code. That holds on any compiler.

The `ntz()` one is in `src/crypto/ocb_internal.cc`, which is Ted Krovetz's OCB v3 reference code — the `#if _MSC_VER` branches came vendored with it, and its own header says "This code is tested for C99 and recent versions of GCC and MSVC". So this isn't a request to add MSVC support: the MSVC branch already in the tree passes an `unsigned*` where `_BitScanForward` takes an `unsigned long*`, and cannot compile as written. If you'd rather not carry code for a compiler you don't support, deleting those branches is an equally good answer and I'll happily redo it that way — or just drop that hunk and keep the header fix.

On the use case: I'm building mosh-client natively on Windows. That's a larger piece of work and I didn't want to lead with it here; these two are the parts that stand on their own.

vmscvx:
Thanks for the edit — yes, a native mosh-client on Windows is where this is going.

On this PR specifically, the two changes aren't the same kind of thing, and my title obscured that.

`swrite.h` has nothing to do with MSVC. The header declares a function taking a `ssize_t` without including anything that defines one; it builds today only because every current includer happens to pull in `<unistd.h>` first. Adding `<sys/types.h>` makes it self-contained and changes no generated code.

`ntz()` is in `src/crypto/ocb_internal.cc`, Ted Krovetz's OCB v3 reference code — the `#if _MSC_VER` branches came vendored with it, and its header says the code is "tested for C99 and recent versions of GCC and MSVC". So this isn't asking you to support MSVC: the MSVC branch already in the tree passes an `unsigned*` where `_BitScanForward` wants an `unsigned long*`, and cannot compile as written. Deleting those branches instead would be just as good an answer — say the word and I'll redo it that way, or drop that hunk and keep only the header fix.

On review capability: that seems to me the real question, and I'd rather answer it with automation than with your time. If it would help, I can add a Windows job to `ci.yml` that builds mosh-client with MSVC and runs the unit tests, so the build is checked by CI rather than by someone having to know Windows.

The POSIX side is meant to need no Windows knowledge to review: every change sits behind `#ifdef _WIN32`, the new files compile to nothing elsewhere, and `make check` and the clang-format 14 lint come out exactly as they do on master — I run both on Linux before pushing.

The rest of the work is split into small branches for the same reason: each is meant to be reviewable, or refusable, on its own.

